PR #11, opened 2016-03-15
[CSV-123] Add possibility to use ResultSet header meta data as CSV header

shivakrishnaah:
CSV-123 added the ResultSet metadata feature using a boolean parameter

https://issues.apache.org/jira/browse/CSV-123


garydgregory:
@shivakrishnaah 
Thank you for you PR> I did this factored in a different way. Please see git master :-)

PR #13, opened 2016-08-23
Fix outdated comments about FileReader

mprudhom:
The comments mention that the constructor that takes a File or Path internally creates a FileReader with the JVM's default CharSet. This is not the case: it instead creates an InputStreamReader with the specified charset. I've removed the incorrect comments.


garydgregory:
The patch did not apply cleanly to our Git master. I made the changes nonetheless. Please verify and close.

mprudhom:
LGTM!

PR #15, opened 2017-02-01
Proposal: New option to distinguish null from empty string

t-suwa:
It is critical to distinguish null from empty string for accomplishing data migration between different RDBMS using csv files.

Unfortunately, CSVParser does not support this functionality and returns ["", "", "", ""] as a result of parsing ",,\\"\\",". Thus, it impossible to make nullable columns have null instead of empty string when importing data from csv files.

So I'd like to add a new option to distinguish them.

- CSVFormat.withMissingColumnValuesAreNull()

With this option on, you can get [null, null, "", null] as a result of above input data. Its default value is off and CSVParser keeps backward compatibility.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -232,7 +232,7 @@ public final class CSVFormat implements Serializable {
      * @see Predefined#Default
      */
     public static final CSVFormat DEFAULT = new CSVFormat(COMMA, DOUBLE_QUOTE_CHAR, null, null, null, false, true, CRLF,
-            null, null, null, false, false, false, false, false);
+            null, null, null, false, false, false, false, false, false);
 
     /**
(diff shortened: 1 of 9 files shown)


garydgregory:
What about:


  Open Declaration   CSVFormat org.apache.commons.csv.CSVFormat.withNullString(String nullString)


Returns a new CSVFormat with conversions to and from null for strings on input and output. 
•Reading: Converts strings equal to the given nullString to null when reading records.
•Writing: Writes null as the given nullString when writing records.
Parameters:nullString the String to convert to and from null. No substitution occurs if nullReturns:A new CSVFormat that is equal to this but with the specified null conversion string.

?

t-suwa:
Not sufficient. It does not provide a way to detect unquoted empty strings only. Take a look at the test case.

https://github.com/t-suwa/commons-csv/commit/9412e001d2e3d598edb6f51e051d2ef0116ad514

or am I missing something ?

garydgregory:
I think this is now obsolete in git master, see commit ac280e705d99cc1168f93f7d9a5f92618bebbb97.

PR #17, opened 2017-03-19
CSV-203: withNullString value is printed without quotes when QuoteMod…

kparoth:
…e.ALL is specified

In my opinion this is only a small change. If I already set the option (QuoteMode.ALL), then it should also be used.

[JIRA-Reference](https://issues.apache.org/jira/browse/CSV-203)

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -38,7 +38,8 @@
     <title>Release Notes</title>
   </properties>
   <body>
-    <release version="1.5" date="2016-MM-DD" description="Bug fix release">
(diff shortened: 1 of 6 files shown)


garydgregory:
But then you cannot do a round trip. How would you differentiate a null value from a string with the same representation?

kparoth:
The programmer should select a value that is unique and does not correspond to one of the possible values. But "N/A" is already quite clear.

garydgregory:
For backward compatibility, maybe we need to add a QUOTE_ALL_NON_NULL

garydgregory:
Closing. see git master.

PR #18, opened 2017-06-25
PROPOSAL: Lines containing whitespace only should not generate empty records

attil-io:
With the following settings, I would expect that any line containing white space only, does NOT generate an additional empty record:

`IgnoreEmpty=true IgnoreSpaces=true`

Reason is, that since white spaces (like tab, space) are ignored, then white-space-only line should count as empty lines (and empty lines are ignored too).

Please tell me if you agree with the idea or not (if you do, I will be happy to do all the necessary steps, like creating JIRA account, opening ticket, signing the CLA, etc.)

I'm also aware, that maybe my code change is not the best solution for the problem (e.g. now ignoreWhiteSpaces is checked twice in `Lexer.nextToken()`), so please treat this PR rather as a POC, and not the final solution. I would be happy to rework it, if you have any suggestions/comments.


garydgregory:
Closing this PR as we have the following:

- org.apache.commons.csv.CSVFormat.withIgnoreEmptyLines()
- org.apache.commons.csv.CSVFormat.withIgnoreEmptyLines(boolean)

PR #20, opened 2017-08-10
CSV-214: adding cache for the line ending information

nmahendru:
Description copied here from the JIRA ticket.
In my use-case , I have to read a CSV file, Mangle some columns and then write out a new csv file with those mangled columns.
I have gone through the parser code and I have found no usable way of getting the line ending information from the CSVParser object. The function readEndOfLine just consumes End of line whether it is CRLF or LF.
Now that problem is that when I am writing my file back using the CSVPrinter, I need to know what line ending my input file was. I could write an external function to do that. A better way would be to store that information in the CSVParser object and just use it.
To do that I am just saving the state in a variable and exposing that using a getter.
I have done some basic testing. and Submitting a pull request regarding that.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVParser.java
+++ b/src/main/java/org/apache/commons/csv/CSVParser.java
@@ -358,7 +358,16 @@ public final class CSVParser implements Iterable<CSVRecord>, Closeable {
         this.characterOffset = characterOffset;
         this.recordNumber = recordNumber - 1;
     }
-
+    /**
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
Thanks for the pull request!

Merged by @garydgregory  in https://github.com/apache/commons-csv/commit/aae6f90442ca09e2461e766a987b33316d9fa6be

It would be nice if you could close the pull request.

Thanks!

nmahendru:
thanks for accepting it.

PR #24, opened 2017-10-08
CSV-217 add flush setting to CSVFormat and create close method with flush setting

alex270295:


alex270295:
Haven't got any good ideas to test `flush` method here. Maybe except this one:
mock `Appendable out` in `CSVPrinter` and verify that flush method was invoked or not depends on flush property value. 

Any suggestions?

garydgregory:
Mocking is likely the only way to go here to make sure the API is called or
not.

Gary

alex270295:
Maybe we should protect or explicitly document behavoiur if both `org.apache.commons.csv.CSVPrinter#close(boolean)`  `org.apache.commons.csv.CSVFormat#withAutoFlush` 
used at same time. Now it will flush if at least one was true.

alex270295:
it seems the new method `org.apache.commons.csv.CSVPrinter#close(boolean)` is not such a cool idea, very badly combined with try-with-resources

src/main/java/org/apache/commons/csv/CSVPrinter.java
```diff
@@ -81,6 +81,17 @@ public CSVPrinter(final Appendable out, final CSVFormat format) throws IOExcepti
 
     @Override
     public void close() throws IOException {
+        close(false);
+    }
+
+    public void close(boolean flush) throws IOException {
+        if (flush || format.getAutoFlush()) {
+            System.out.println(flush);
```
garydgregory: Don't need those println()'s ;-)

alex270295: sry, debug =)

PR #25, opened 2018-02-09
CSV-216: Add mutator withValue() to CSVRecord

stain:
Addresses [CSV-215](https://issues.apache.org/jira/browse/CSV-215) and [CSV-216](https://issues.apache.org/jira/browse/CSV-216)

I know mutable `CSVRecord` caused [quite](https://lists.apache.org/thread.html/8f038aac6173d8be63b14c612c6596a613a2c8f0b16de446ea732885@%3Cdev.commons.apache.org%3E) a [discussion](https://lists.apache.org/thread.html/57478abc7f4947e7fe61fe9e46770faf888df156eabdc2fbc6c0c836@%3Cdev.commons.apache.org%3E) earlier.

@garydgregory created a CSV-216 branch b23f963e8dd4ca553a233e653a6220d8e80db9e9 adding a new `CSVMutableRecord` which could be enabled in the format `withMutableRecords(true)`

I like that this means the records remain immutable unless explicitly turned on (hence a `CSVRecord` would be safe to pass around), but I had trouble with this as that required casting to `CSVMutableRecord` to actually mutate.

It is also a bit weird to only be able to mutate a record in the middle of a parsing iterator  - rows are still "immutable".  There is no way to remove/reorder/insert records other than doing it out of bands
(e.g. construct a new List).

This pull request augments the branch to add mutator functions:

```java
for (CSVRecord r : csvparser) {
  CSVRecord rSoup = r.withValue(4, "soup")
                         .withValue(5, "fish");
  // original r is untouched and can be used again
  CSVRecord rBeans = r.withValue(3, "beans");
  someList.add(r);
  someList.add(rSoup);
  someList.add(rBeans);
}
```

By default the records from the parser are immutable (`.isMutable()` return `false`), and the `with` functions create a clones on demand. The mutated records are mutable, but you can force either with `.mutable()` or `.immutable()` -- which would either clone or return itself depending on `isMutable()`.  

Setting `.withMutableRecords(true)` in the format can be used to avoid the initial array clone of `.withValue`.

The subclass `CSVMutableRecord` is now package-private and just short-circuits some of the above, as the mutator functions are always available.

For completeness I added `.withComment()` -- but I'm not sure about the current approach here as it means I have to make `CSVRecord.comment` non-final and package-accessible to `CSVMutableRecord` (or it would have to keep a duplicate `comment` field and override all the comment methods)

(This would have been easier if `CSVRecord` was an interface and we could avoid subclassing.)

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVMutableRecord.java
+++ b/src/main/java/org/apache/commons/csv/CSVMutableRecord.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.commons.csv;
+
+import java.util.Map;
+
+public final class CSVMutableRecord extends CSVRecord {
+
+    private static final long serialVersionUID = 1L;
+
+    CSVMutableRecord(String[] values, Map<String, Integer> mapping, String comment, long recordNumber,
+            long characterPosition) {
+        super(values, mapping, comment, recordNumber, characterPosition);
+    }
+
+    @Override
+    public void put(int index, String value) {
+        super.put(index, value);
+    }
+
+    @Override
+    public void put(String name, String value) {
+        super.put(name, value);
+    }
+
+}
--- a/src/main/java/org/apache/commons/csv/CSVMutableRecord.java
+++ b/src/main/java/org/apache/commons/csv/CSVMutableRecord.java
@@ -19,7 +19,19 @@ package org.apache.commons.csv;
 
 import java.util.Map;
 
-public final class CSVMutableRecord extends CSVRecord {
+/**
+ * A mutable version of CSVRecord
+ * <p>
+ * As mutation is generally done within the parent {@link CSVRecord}, this 
+ * package-private class just includes more efficient versions of 
+ * mutator functions, bypassing the need to make copies 
+ * (except in {@link #immutable()}).
+ * <p>
+ * To enable generating CSVMutableRecord by default, set 
+ * {@link CSVFormat#withMutableRecords(boolean)} to <code>true</code>.
+ *
+ */
+final class CSVMutableRecord extends CSVRecord {
 
     private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/apache/commons/csv/CSVRecord.java
```diff
@@ -272,5 +337,52 @@ public String toString() {
...
+     * @throws IllegalStateException
+     *             if no header mapping was provided
+     * @throws IllegalArgumentException
+     *             if {@code name} is not mapped or if the record is inconsistent
+     * @return A mutated CSVRecord
+     */
+    public CSVRecord withValue(String name, String value) {    	
+    	CSVRecord r = mutable();
+    	r.put(name,  value);
+    	return r;
```
stain: Note that here the javadoc leaves mutability of the returned CSVRecord undefined, this implementation will in fact always return a mutable CSVRecord (as can be verified with `isMutable()`)

If someone did `record.withValue(1, "b").immutable()` it would cause two array copies. But `record.withValue(1, "b").withValue(2, "c") would just make one array copy (or 0 if withMutableRecords was set in format).

src/main/java/org/apache/commons/csv/CSVMutableRecord.java
```diff
@@ -0,0 +1,65 @@
...
+ * <p>
+ * To enable generating CSVMutableRecord by default, set 
+ * {@link CSVFormat#withMutableRecords(boolean)} to <code>true</code>.
+ *
+ */
+final class CSVMutableRecord extends CSVRecord {
+
+    private static final long serialVersionUID = 1L;
+
+    CSVMutableRecord(String[] values, Map<String, Integer> mapping, String comment, long recordNumber,
```
stain: I am considering if there's not much point in the `CSVMutableRecord` class anymore, except in any hope of the compiler being clever due to the final methods and the inline `isMutable()`  constant.

An alternative is to add a `private final boolean mutable` to `CSVRecord` and make it final again.

src/main/java/org/apache/commons/csv/CSVRecord.java
```diff
@@ -203,18 +238,48 @@ public boolean isSet(final String name) {
...
+    public final CSVRecord mutable() {
+    	if (isMutable()) {
+    		return this;
+    	}
+		String[] newValues = Arrays.copyOf(values, values.length);
+    	return new CSVMutableRecord(newValues, mapping, comment, recordNumber, characterPosition);
+	}    
+
+    final void put(final int index, String value) {
+        values[index] = value;
```
stain: Should it be possible to modify columns outside the length of `values`?  (Would cause multiple array resizes if done many times)

stain:
f66a839 adds warnings about `.clone()` according to [discussion](https://lists.apache.org/thread.html/e7c54a595325f1faf7c78b86e76c048c98a914fbb869b8f6d648e4f5@%3Cdev.commons.apache.org%3E) with Gilles @sebbASF et al.

Is this OK to merge? @garydgregory said he would have a look.

garydgregory:
I would to review today or Thursday.

stain:
Thanks, @garydgregory -- let me know if you feel this is not mature enough; I can always leave this PR out and run a 1.3 RC as-is.

garydgregory:
Hello @stain and all,

Starting in version 1.10.0, you can use the method CSVRecord.values() and do whatever you want with the array.

This might avoid the need for this PR.

garydgregory:
@stain ping.

garydgregory:
Closing , no reply in 6+ months.

PR #29, opened 2018-05-21
CSV-222 ignore set of characters

LostKatana:
This should solve the problem when the CSV to be read contains invalid characters.
It reverences to the issue https://issues.apache.org/jira/browse/CSV-222


LostKatana:
Closed as easier solution is proposed.

PR #32, opened 2018-10-07
[CSV-208] Fix escape character for POSTGRESQL_TEXT and POSTGRESQL_CSV formats

Jurrie:


garydgregory:
You patch seems to contradict https://www.postgresql.org/docs/current/static/sql-copy.html which states:

```
QUOTE

    Specifies the quoting character to be used when a data value is quoted. The default is double-quote. This must be a single one-byte character. This option is allowed only when using CSV format.
ESCAPE

    Specifies the character that should appear before a data character that matches the QUOTE value. The default is the same as the QUOTE value (so that the quoting character is doubled if it appears in the data). This must be a single one-byte character. This option is allowed only when using CSV format.
```

WDYT?

What is the exact COPY TO command you used to export you data?

Jurrie:
Hi @garydgregory,

I did not use the COPY TO command to export anything, I'm importing a dataset that is exported by another company for me in CSV. The source is not PostgreSQL.

Indeed, the PostgreSQL documentation seems to suggest differently than the Javadoc above the `POSTGRESQL_CSV` and `POSTGRESQL_TEXT` constants. I will need to investigate by using COPY TO to actually export something containing a double quote.

However, I think there is another issue. Commons-csv can not handle the situation where the quote character and the escape character are the same. My dataset does not contain a double quote that should be escaped (in other words: the double quotes in the set are all quote characters, none are escape characters). But still, Commons-csv fails to read the CSV file.

As I figured that the escape character for the PostgreSQL formats was wrong, I did not investigate the quote char == escape char bug any further.

garydgregory:
How does your latest change match the documentation? ESCAPE says the "default is the same as the QUOTE value" and QUOTE says "The default is double-quote." Where are you seeing backslash?

Jurrie:
PostgreSQL's COPY command allows for multiple formats.

One is [CSV format](https://www.postgresql.org/docs/9.5/static/sql-copy.html#AEN74952). The [documentation on the parameters QUOTE and ESCAPE](https://www.postgresql.org/docs/9.5/static/sql-copy.html#AEN74634) says both parameters are `"` by default, and both are allowed only when using CSV format. So that clears up what the `POSTGRESQL_CSV` format in Commons-csv should be: quoute char should be `"` and escape char should be `"`. This was already the case; only the accompanying Javadoc was wrong.

Another format is [TEXT](https://www.postgresql.org/docs/9.5/static/sql-copy.html#AEN74869). The documentation on the TEXT format reads: "Backslash characters (\\) can be used in the COPY data to quote data characters that might otherwise be taken as row or column delimiters. In particular, the following characters must be preceded by a backslash if they appear as part of a column value: backslash itself, newline, carriage return, and the current delimiter character."

I tested this by doing `copy (select 'col1'||chr(9)||'end',  'col2'||chr(10)||'end') to '/tmp/test.txt' (format text);`. The file reads `col1\tend       col2\nend`. So that leads me to believe that the escape char in `POSTGRESQL_TEXT` format should be `\`, like the Javadoc already says.

garydgregory:
Thank you for your explanation. Please verify what it in git master now.

garydgregory:
No reply in years, closing.

PR #33, opened 2018-10-10
Performance Enhancement: Call toArray with Zero Array Size

DaGeRe:
toArray should, at least since Java 7, be used with new T[0] instead of size, since it is "faster, safer, and contractually cleaner" (https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion)


garydgregory:
@DaGeRe 
Thank you for the PR. I did this slightly differently in git master by reusing a constant. Closing.

PR #39, opened 2019-02-19
[CSV-238] Escape quotes in CLOBs

luyseyal:
(The bad coveralls report is from code merged to master this last weekend, not this patch)

This fixes Excel output by enabling the escaping of quotes in CLOBs.

Given a CLOB like:
```
Choose either
"Y" or "N"
```
The patch will transform it to:
```
Choose either
""Y"" or ""N""
```


luyseyal:
@garydgregory

luyseyal:
@britter  Can you review this? Gary's code in master does fix the general idea behind outputting CLOBs but it doesn't escape quotes embedded in the CLOB string. This patch enables escaping the quotes so that Excel output will work (and probably others).

PR #40, opened 2019-05-05
added feature: pretty printer utility

ali-ghanbari:
For artifact evaluation of our ISSTA paper we needed a pretty printer for CSV files.

I am hereby presenting a crude implementation of a pretty printer utility that, in my humble opinion, is good to be part of Apache Commons CSV.

I will be more than happy to improve my implementation based on your comments.

### DIFF
--- a/src/main/java/org/apache/commons/csv/pretty/PrettyPrinter.java
+++ b/src/main/java/org/apache/commons/csv/pretty/PrettyPrinter.java
@@ -1,0 +1,172 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.commons.csv.pretty;
+
+import org.apache.commons.csv.CSVParser;
+import org.apache.commons.csv.CSVRecord;
+import org.apache.commons.lang3.StringUtils;
+
+import java.io.IOException;
+import java.io.PrintStream;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Pretty printing utility
+ *
+ * @author Ali Ghanbari
+ */
+public class PrettyPrinter {
+    protected final List<CSVRecord> records;
+    protected final TableStyle style;
+    protected final PrintStream out;
+    protected final int[] spaces;
+
+    public PrettyPrinter(final List<CSVRecord> records, final TableStyle style, final PrintStream out) {
+        this.records = records;
+        this.style = style;
+        this.out = out;
+        this.spaces = calculateSpaces(records);
+    }
+
+    public PrettyPrinter(final CSVParser parser, final TableStyle style) throws IOException {
+        this.records = parser.getRecords();
+        this.style = style;
+        this.out = System.out;
+        this.spaces = calculateSpaces(this.records);
+    }
+
+    public PrettyPrinter(final CSVParser parser, final TableStyle style, final PrintStream out) throws IOException {
+        this.records = parser.getRecords();
+        this.style = style;
+        this.out = out;
+        this.spaces = calculateSpaces(this.records);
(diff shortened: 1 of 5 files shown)


garydgregory:
I would have thought something like this would be fit better in our framework as a `CSVFormat`.

src/main/java/org/apache/commons/csv/pretty/TableStyle.java
```diff
@@ -0,0 +1,220 @@
...
+ * We have prepared two pre-defined styles:
+ * <ul>
+ *     <li><code>SIMPLE</code> uses simple ASCII characters to draw the table</li>
+ *     <li><code>FANCY</code> uses so-called Extended ASCII characters to draw fancier tables</li>
+ * </ul>
+ *
+ *
+ * @author Ali Ghanbari
+ */
+public final class TableStyle implements Serializable {
```
garydgregory, line 34: I believe this should be an enum. Why make this an enum BTW?

src/test/java/org/apache/commons/csv/pretty/PrettyPrinterTest.java
```diff
@@ -0,0 +1,129 @@
...
+import java.net.URL;
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Ali Ghanbari
+ */
+public class PrettyPrinterTest {
```
garydgregory, line 36: This test does not assert anything, so it will fail to detect regressions.

src/main/java/org/apache/commons/csv/pretty/TableStyle.java
```diff
@@ -0,0 +1,220 @@
...
+            '┌',
+            '╔',
+            '┐',
+            '╗',
+            '└',
+            '╚',
+            '┘',
+            '╝');
+
+    public final char THIN_HORIZONTAL_LINE;
```
garydgregory, line 87: Instance variables should be in lower-case and camel-cased, NOT in upper case, for example `THIN_VERTICAL_LINE` should be `thinVertialLine`. Also, public instance variables break encapsulation.

src/main/java/org/apache/commons/csv/pretty/TableStyle.java
```diff
@@ -0,0 +1,220 @@
...
+    public final char THIN_UPPER_LEFT_CORNER_JOINT;
+    public final char THICK_UPPER_LEFT_CORNER_JOINT;
+    public final char THIN_UPPER_RIGHT_CORNER_JOINT;
+    public final char THICK_UPPER_RIGHT_CORNER_JOINT;
+    public final char THIN_LOWER_LEFT_CORNER_JOINT;
+    public final char THICK_LOWER_LEFT_CORNER_JOINT;
+    public final char THIN_LOWER_RIGHT_CORNER_JOINT;
+    public final char THICK_LOWER_RIGHT_CORNER_JOINT;
+
+    public TableStyle(final char THIN_HORIZONTAL_LINE,
```
garydgregory, line 112: Same comment as for instance variable: lower-case and camel-case.

src/main/java/org/apache/commons/csv/pretty/PrettyPrinter.java
```diff
@@ -0,0 +1,172 @@
...
+
+    public void print(final boolean thickBorder,
+                      final boolean distinguishHeader,
+                      final boolean distinguishRows,
+                      final boolean cellBorder) {
+        final Iterator<CSVRecord> recordIterator = this.records.iterator();
+        if (!recordIterator.hasNext()) {
+            return;
+        }
+        printUpperLine(thickBorder, cellBorder ? (thickBorder ? this.style.THICK_DOWNWARD_JOINT : this.style.THIN_DOWNWARD_JOINT) : this.style.THICK_HORIZONTAL_LINE);
```
garydgregory, line 152: Changing "Upper/Lower" to "Start/End" is better IMO.

garydgregory:
I've sprinkled comments throughout. I am not sure about how this would fit as a CSVFormat subclass.

src/main/java/org/apache/commons/csv/pretty/PrettyPrinter.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+        out.print(sb.substring(0, sb.length() - 1));
+        out.print(thickBorder ? (thickLine ? this.style.THICK_THICK_RIGHT_JOINT : this.style.THICK_RIGHT_JOINT) : this.style.THIN_RIGHT_JOINT);
+    }
+
+    private void printLowerLine(final boolean thick) {
+        if (this.spaces.length == 0) {
+            return;
+        }
+        out.print(thick ? this.style.THICK_LOWER_LEFT_CORNER_JOINT : this.style.THIN_LOWER_LEFT_CORNER_JOINT);
```
garydgregory, line 117: I thought the point of the style class what to abstract out this constant testing of "thick". The constant testing of "thick" seems like a code smell and not very OO...

garydgregory:
Closing: no reply in over a year.

PR #45, opened 2019-06-20
adding power support

ghatwala:
to ensure this repo builds on ppc64le , added support here

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@
 # limitations under the License.
 
 language: java
+os:
+  - linux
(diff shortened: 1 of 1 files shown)


aherbert:
This is Java and as such the platform should be considered the JDK and not the underlying OS.

Unless the codebase uses functionality not part of the JDK such as custom JNI calls that are platform dependent this change is really a test of the functionality of the JDK. Testing a vendor implementation of the JDK is outside the scope of commons.

garydgregory:
CLosing power vs. not should not matter.

PR #47, opened 2019-07-03
CSV-247: CSVParser to check an empty header before checking duplicates.

aherbert:
This rearranges the logic of checking for empty column headers before checking for duplicates.

I've left the rest of the logic alone. The implication is that it will always allow a duplicate header if the header is empty irrespective of whether duplicates are not allowed. So you can configure the parser to not allow duplicates but allow missing headers and then it skips the fact that each missing header is a duplicate. If documented this in the code.

I also note that the header is checked for a duplicate without trim but the empty header check uses trim. This is contradictory but perhaps a user does want to have headers `"A"," A "," A"`. Any proper CSV header would not whitespace pad headers so I'll leave this until someone reports it as a problem. 

The fix found that the existing test `CSVParserTest.testHeaderMissing()` did not set the AllowMissingColumnNames property. If this test is to work with a missing header then it needs that setting.

I also updated the tests of missing headers which use 5 columns to have 5 entries in the record. Previously it was 4 which is not correct.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVParser.java
+++ b/src/main/java/org/apache/commons/csv/CSVParser.java
@@ -495,19 +495,18 @@ public final class CSVParser implements Iterable<CSVRecord>, Closeable {
             if (headerRecord != null) {
                 for (int i = 0; i < headerRecord.length; i++) {
                     final String header = headerRecord[i];
-                    final boolean containsHeader = header != null && hdrMap.containsKey(header);
                     final boolean emptyHeader = header == null || header.trim().isEmpty();
-                    if (containsHeader) {
-                        if (!emptyHeader && !this.format.getAllowDuplicateHeaderNames()) {
-                            throw new IllegalArgumentException(
-                                String.format(
-                                    "The header contains a duplicate name: \"%s\" in %s. If this is valid then use CSVFormat.withAllowDuplicateHeaderNames().",
-                                    header, Arrays.toString(headerRecord)));
-                        }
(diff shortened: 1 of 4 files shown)


mykolapolonskyi:
hi there 
I've found you PR related to the thing that I've catched - when csv file doesn't contain the header line I parsing should throw IllegalArgument Exception 

Maybe this test can be added as part of your PR because seems that fix should be in place that you`ve edited? what do you think?

```
private enum CSVHeaders {
        Name;
}
@Test(expected = IllegalArgumentException.class)
public void testMissingHeaderWithEnumConfiguration() throws Exception {
final Reader in = new StringReader("not name");
CSVFormat.DEFAULT.withHeader(CSVHeaders.class).withIgnoreHeaderCase().parse(in).iterator();
}
```

aherbert:
Hi @mykolapolonskyi .

I do not understand your bug in your demo code. Are you testing that if the format has an enumeration of header names and you parse an input that does not match the header it should throw an exception?

IIUC the withHeader(Enum) method is used to define an array of headers for the format. The parse method will then read records and skip trying to read a header because it has one.

I don't think your issue is related to this PR as this addresses code to parse the header from the input. With your example you are skipping this code as you provide a header. Please ask on the [Commons CSV user mailing lists](https://commons.apache.org/proper/commons-csv/mail-lists.html) about this. It may be an incorrect use of the functionality or it may be a bug. If so it should be addressed in a separate issue.

PR #49, opened 2019-10-05
CSV-252: Migration to JUnit Jupiter

mureinik:
This PR upgrades the project's testing framework from JUnit 4.12 to the modern JUnit Jupiter 5.5.4.

Since JUnit 5 Jupiter is not backwards compatible to JUnit 4.x (or even JUnit Vintage), this patch is a bit large, even though a lot of the changes are merely cosmetic (such as changing the argument order, see details below). In order to make the reviewer's task as easy as possible, this PR does not presume to use JUnit Jupiter's best practices and all its new functionality, but only to migrate the existing tests with as little change as possible. Following PRs may want to improve the tests by using some of JUnit Jupiter's new features.

This PR includes the following changes:

1. Maven dependency changes:
     1. `junit:junit` was replaced with `org.junit.jupiter:junit-jupiter`.
     1. `org.hamcrest:hamcrest` was introduced as an explicit dependency, since the project uses Hamcrest, and JUnit Jupiter does not bundle Hamcrest, unlike JUnit 4.x.
     1. The Mockito dependency was upgraded from `org.mockito.mockito-all:1.10.19` to `org.mockito:mockito-core:3.1.0`, which supports JUnit Jupiter.

 1. Annotations:
      1. `org.junit.jupiter.api.Test` was used as a drop in replacement for `org.juit.Test` without arguments. See 3.ii. for handling of `@Test` annotations with an `expected` argument.
     1. `org.junit.jupiter.api.BeforeEach` was used as a drop in replacement for `org.junit.Before`.
     1. `org.junit.jupiter.api.BeforeAll` was used as a drop in replacement for `org.junit.BeforeClass`.
     1. `org.junit.jupiter.api.Disabled` was used as a drop in replacement for `org.junit.Ignore`.

1. Assertions:
     1. `org.junit.jupiter.api.Assertions`' methods were used as drop in replacements for `org.junit.Assert`'s methods with the same name in the simple case of an assertion without a message. In the case of an assertion with a message, `org.junit.jupiter.api.Assertions`' methods were used, but the argument order was changed - `Assert`'s methods take the message as the first argument, while `Assertions`' methods take the message as the last argument.
     1. `org.junit.jupiter.api.Assertions#assertThrows` was used to assert that a specific exception was thrown instead of an `org.junit.Test` annotation with an `expected` argument. This technique has a couple of side bonuses. First, it makes the tests slightly stricter, as now they can assert the exception was thrown from a specific line and prevent false positives where the test's "set-up" code accidentally threw that exception. Second, it clarifies that some of the test code is unreachable (as a previous line already throws an exception), and can safely be removed in order to clean up the test. The throws clauses of these methods were cleaned up from exceptions that can no longer be thrown in order to avoid compilation warnings.
     1. `org.hamcrest.MatcherAssert#assertThat` was used as a drop in replacement for `org.junit.Assert#assertThat`.
     1. `org.junit.jupiter.api.Assertions#assertNull` was used as a drop in replacement for `junit.framework.TestCase#assertNull`.

1. Specific Changes:
     1. `CSVFileParserTest` was rewritten with JUnit Jupiter's `org.junit.jupiter.api.ParameterizedTest`.  Unlike JUnit 4's `org.junit.runners.Parameterized`, it cannot be used to inject arguments to a test's construct, and so the test can't be stateful. Instead, it was rewritten so every test receives the file as a parameter, and opens a reader on it itself. As a side bonus, this design makes it easier to close the reader and avoid leaving open file descriptors like the original test did.

Closes CSV-252

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,8 +40,8 @@ CSV files of various types.
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-all</artifactId>
-      <version>1.10.19</version>
+      <artifactId>mockito-core</artifactId>
+      <version>3.1.0</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -33,9 +33,15 @@ CSV files of various types.
 
   <dependencies>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>4.12</version>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <version>5.5.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest</artifactId>
+      <version>2.1</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/src/test/java/org/apache/commons/csv/AssertionsTest.java
+++ b/src/test/java/org/apache/commons/csv/AssertionsTest.java
@@ -17,7 +17,9 @@
 
 package org.apache.commons.csv;
 
-import org.junit.Test;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
 
 /**
  */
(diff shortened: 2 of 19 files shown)


garydgregory:
@mureinik 

Thank you for your contribution to modernize Apache Commons CSV! :-)

Gary

PR #51, opened 2019-10-28
[CSV-253] Handle absent values in input

lbruun:
Being able to appropriately translate an absent value in CSV input with a Java `null` value. Previously, there was no way to do this, such a value would at best become a zero-length string when parsing. This made it impossible to correctly parse CSV output from say databases.

This PR is in reference to [CSV-253](https://issues.apache.org/jira/browse/CSV-253).

The PR addresses the issue by adding a flag on `Token` so that it becomes possible to distinguish between a token which is the result of an absent value in input or an actual zero-length string. A new modifier, `absentIsNull` is introduced on `CSVFormat`.   All existing formats and functionality are kept as-is, meaning the new feature is fully based on opt-in.

As a possible next step the pre-defined CSV formats for databases (i.e. `INFORMIX_UNLOAD_CSV`, `MYSQL`, `ORACLE` and `POSTGRESQL_CSV`) should be reviewed. I suspect that at least `POSTGRESQL_CSV` has always been incorrect in this matter. With this PR it can be corrected (if need be).

I've taken the liberty of adding "since 1.8" in the Javadocs since I see some commits for preparing such a release. Thus, hoping to include this.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -266,7 +266,7 @@ public final class CSVFormat implements Serializable {
      * @see Predefined#Default
      */
     public static final CSVFormat DEFAULT = new CSVFormat(COMMA, DOUBLE_QUOTE_CHAR, null, null, null, false, true, CRLF,
-            null, null, null, false, false, false, false, false, false, true);
+            null, false, null, null, false, false, false, false, false, false, true);
 
     /**
      * Excel file format (using a comma as the value delimiter). Note that the actual value delimiter used by Excel is
@@ -672,7 +672,7 @@ public final class CSVFormat implements Serializable {
      * @see #TDF
      */
     public static CSVFormat newFormat(final char delimiter) {
-        return new CSVFormat(delimiter, null, null, null, null, false, false, null, null, null, null, false, false,
+        return new CSVFormat(delimiter, null, null, null, null, false, false, null, null, false, null, null, false, false,
                 false, false, false, false, true);
     }
 
@@ -711,6 +711,8 @@ public final class CSVFormat implements Serializable {
     private final boolean ignoreSurroundingSpaces; // Should leading/trailing spaces be ignored around values?
 
     private final String nullString; // the string to be used for null values
+
+    private final boolean absentMeansNull;   // if absent values in input should be treated as null
 
     private final Character quoteCharacter; // null if quoting is disabled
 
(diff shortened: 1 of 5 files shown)


divijvaidya:
Hi @lbruun 
Are you planning to work on this PR? I was wondering what was it that is left to do here.

lbruun:
@divijvaidya. The PR is what it is and is complete. However, since time has now passed the PR now has a minor merge conflict. Also my changes to Javadoc states (for this new feature) "since 1.8" which seemed reasonable at the time because it was the the (then) upcoming release and I was hoping the maintainers would include my PR for that release. Version 1.8 has since then been released _without_ this PR so that would need to be changed in the PR.

Anyway, my main problem is that I haven't been able to explain to maintainers of this library why this is a missing feature in the library. In short why the library should be able to distinguish between (when parsing):

```text
545-03-1234,"","Doe"
```

and 
```text
545-03-1234,,"Doe"
```

..and why the existing feature, `org.apache.commons.csv.CSVFormat.nullString`, doesn't address this problem. You can see the explanation and discussion [here](https://issues.apache.org/jira/browse/CSV-253?focusedCommentId=16962105&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16962105).

Note that once maintainer(s) of the library agree to the problem being present then the next discussion is obviously _how_ to address it. My suggestion in this PR favors strict backwards compatibility (no breaking change) : I simply introduce a new attribute rather than changing what the existing attribute `nullString` does. This will probably be confusing for users of the library, but the alternative - breaking change - is worse IMHO.

I'll be glad to update the PR if I thought it had chance of being accepted.

garydgregory:
I am wondering if instead of inventing yet another setting, the parser could/should use `QuoteMode.ALL_NON_NULL` to parse an non-quoted empty string as a `null`. Needs thinking and feedback... Note that `QuoteMode` is currently only used for printing, but it would make sense for us to be able to reuse it.

lbruun:
@garydgregory I think what you are essentially saying is to do [PR 77](https://github.com/apache/commons-csv/pull/77) rather than this one. Either way will work.

It is all about the appetite for breaking changes (PR77) vs a new opt-in feature (this PR). On one hand you can argue that the library should have been able to distinguish between empty string and absent string from the very beginning and therefore fixing it by changing existing behavior is justified, meaning do PR77. But in the world of CSV only a single generally agreed-upon standard exists, RFC4180, and that one is inconclusive about this scenario. We also have the [W3's Recommendation for Tabular Data](https://www.w3.org/TR/tabular-data-model/) but that one actually clearly states that the empty string and absent string are the same. This speaks towards that the way the library currently works should be the default and therefore that a breaking change it not acceptable.

Example: Let's say a user is currently using static format `POSTGRESQL_CSV` to parse output which originates from a PostgreSQL database. Such user will with PR77 potentially start to see NPEs just because he updated to latest version of Apache Commons CSV. Ouch!

However, what speaks in favor of a breaking change, meaning PR77, is that most users will possibly (unfairly?) expect a CSV format to be _lossless_, meaning that if they export from their database into CSV and then re-import that CSV file into same database then no data has been lost or distorted. With PR77 you uphold the principle of lossless conversion and if this is what users implicitly expect from the library then PR77 is the way to go.

Personally I think I'm leaning towards PR77, even if it is a breaking change. It just needs clear documentation,

garydgregory:
commit ac280e705d99cc1168f93f7d9a5f92618bebbb97

PR #52, opened 2019-12-25
Add CSVRecord.isSet(int) method

0x100:
Added the isSet(int) method to CSVRecord for checking whether a column with a given index is set. Also removed unnecessary unboxing of an Integer variable in the CSVRecord#isSet(String) method.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVRecord.java
+++ b/src/main/java/org/apache/commons/csv/CSVRecord.java
@@ -215,6 +215,17 @@ public final class CSVRecord implements Serializable, Iterable<String> {
     }
 
     /**
+     * Checks whether a column with given index has a value.
+     *
+     * @param i
+     *         a column index (0-based)
+     * @return whether a column with given index has a value
+     */
+    public boolean isSet(final int i) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/csv/CSVRecord.java
```diff
@@ -211,7 +211,18 @@ public boolean isMapped(final String name) {
      * @return whether a given columns is mapped and has a value
      */
     public boolean isSet(final String name) {
-        return isMapped(name) && getHeaderMapRaw().get(name).intValue() < values.length;
+        return isMapped(name) && getHeaderMapRaw().get(name) < values.length;
```
garydgregory: Let's keep the boxing as is please and focus this PR on the new API. 

If you want to go through the whole code base and consistently use one boxing style that's a different PR IMO. But: I prefer to know when boxing and unboxing takes place since it is not free; this gives the compiler the opportunity to flag unintentional boxing and unboxing and let the programmer decide that's really what should happen.

0x100: Ok, I've rolled back this change.

PR #54, opened 2020-01-17
Add testcases for CSVRecord with get(Enum) and toString.

dota17:


garydgregory:
@dota17 
May you please resolve conflicts? I'm afraid master's diverged enough since you created this PR.

PR #61, opened 2020-03-13
[CSV-259] CSVFormat.printWithEscapes throws StringIndexOutOfBoundsException when value is Reader

dota17:
fix isssue csv-259 [https://issues.apache.org/jira/browse/CSV-259](url)
printWithEscapes gets StringIndexOutOfBoundsException when value is Reader


garydgregory:
@dota17 
For this PR to be properly validated, you will need to add a test that fails without the change in main.

dota17:
@garydgregory add a testcase

dota17:
@garydgregory hi, gary, is still any questions about this pr? I can fix this pr. 
And I have add a new testcase and fix some codes for CSV246 with some changes for pom.xml too . So I couldn't create a new pr wihtout this merged.

garydgregory:
The link to https://issues.apache.org/jira/browse/CSV-259 is broken so here is a `real` one.

PR #62, opened 2020-03-18
Fix CSVFileParserTest.java to allow for a null return value from record.getComment()

dota17:
The old test case record.getComment() will never be null and if record.getComment() be null the test code misplace the null test.
Add a new test file that record.getComment() will be null and test record.getComment() no null before using.

### DIFF
--- a/src/test/java/org/apache/commons/csv/CSVFileParserTest.java
+++ b/src/test/java/org/apache/commons/csv/CSVFileParserTest.java
@@ -93,8 +93,9 @@ public class CSVFileParserTest {
             try (final CSVParser parser = CSVParser.parse(new File(BASE, split[0]), Charset.defaultCharset(), format)) {
                 for (final CSVRecord record : parser) {
                     String parsed = Arrays.toString(record.values());
-                    if (checkComments && record.getComment() != null) {
-                        parsed += "#" + record.getComment().replace("\n", "\\n");
+                    String comment = record.getComment();
+                    if (checkComments && comment != null) {
+                        parsed += "#" + comment.replace("\n", "\\n");
                     }
                     final int count = record.size();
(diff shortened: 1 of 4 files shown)


src/test/java/org/apache/commons/csv/CSVFileParserTest.java
```diff
@@ -93,11 +93,8 @@ public void testCSVFile(final File testFile) throws Exception {
             try (final CSVParser parser = CSVParser.parse(new File(BASE, split[0]), Charset.defaultCharset(), format)) {
                 for (final CSVRecord record : parser) {
                     String parsed = Arrays.toString(record.values());
-                    if (checkComments) {
-                        final String comment = record.getComment().replace("\n", "\\n");
-                        if (comment != null) {
```
garydgregory: Let's keep the caching of "record.getComment()"

src/test/java/org/apache/commons/csv/CSVFileParserTest.java
```diff
@@ -140,11 +137,8 @@ public void testCSVUrl(final File testFile) throws Exception {
             try (final CSVParser parser = CSVParser.parse(resource, Charset.forName("UTF-8"), format)) {
                 for (final CSVRecord record : parser) {
                     String parsed = Arrays.toString(record.values());
-                    if (checkComments) {
-                        final String comment = record.getComment().replace("\n", "\\n");
-                        if (comment != null) {
-                            parsed += "#" + comment;
-                        }
+                    if (checkComments && record.getComment() != null) {
```
garydgregory: Same as above.

src/test/java/org/apache/commons/csv/CSVFileParserTest.java
dota17: thanks for review , I have fix that

PR #68, opened 2020-05-08
[CSV-150] Escaping is not disableable

dota17:
* https://issues.apache.org/jira/projects/CSV/issues/CSV-150

Hi, @garydgregory 
In the source code, it is estimated that for performance reasons, ```Character``` is converted to ```char```, so ```null``` is mapped to ```\ufffe``` which cannot appear in the middle of a normal file. Now if you have the ```\ufffe``` character in the middle of the file, there's a problem.

### DIFF
--- a/src/test/java/org/apache/commons/csv/issues/JiraCsv150Test.java
+++ b/src/test/java/org/apache/commons/csv/issues/JiraCsv150Test.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.csv.issues;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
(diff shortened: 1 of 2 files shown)


garydgregory:
@dota17 
May you please rebase on master?

garydgregory:
@dota17 
May you please rebase on master?

huxi:
Looks like the tests doesn't fail in case of `IOException`.

src/test/java/org/apache/commons/csv/issues/JiraCsv150Test.java
```diff
@@ -0,0 +1,64 @@
...
+        } finally {
+            if (csvParser != null) {
+                csvParser.close();
+            }
+        }
+    }
+
+    @Test
+    public void testDisableEscaping() throws IOException {
+        final StringReader stringReader = new StringReader("\"66\u2441\",,\"\",\"DeutscheBK\ufffe\",\"000\"\r\n");
```
garydgregory, line 49: Hi @dota17 
This feels like a made-up use case (not real-life), since https://codepoints.net/U+FFFE?lang=en says this should not happen. 

Are you saying that this character occurs in a "real" data file? How is this file produced?

huxi: Hi @garydgregory 
I'm not the author of [CSV-150](https://issues.apache.org/jira/projects/CSV/issues/CSV-150) or this pull request but I can assure you that I commented above because I had this issue in real data.

This happens because everything is terrible, always. The use case is cleansing of malformed data.

According to [Wikipedia](https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Non-characters), the expected behavior changed.

> Versions of the Unicode standard from 3.1.0 to 6.3.0 claimed that noncharacters "should never be interchanged". [Corrigendum #9](https://www.unicode.org/versions/corrigendum9.html) of the standard later stated that this was leading to "inappropriate over-rejection", clarifying that "[Noncharacters] are not illegal in interchange nor do they cause ill-formed Unicode text", and removing the original claim.

garydgregory: @huxi 
Thank you for the pointers! I'll take a look.

garydgregory:
@dota17, @huxi 
Fixed differently in git master.

PR #69, opened 2020-05-08
Change JiraCsv164Test to JiraCsv154Test

dota17:


dota17:
The content of this file is the test case of CSV-154. So the name should be modified to JiraCsv154Test.

garydgregory:
@dota17 Please stop creating PRs while leaving the description field empty. Reviewed should not have to parse out your changes to try and understand why a PR is needed. In this specific case, the description should state _why_ you are making this change? Has the test code's purpose changed? Was it never called the right name in the first place? Or, was the Jira ticket itself changed?

garydgregory:
> The content of this file is the test case of CSV-154. So the name should be modified to JiraCsv154Test.

Then say that in the description!

dota17:
@garydgregory 
Oh sorry, I will pay attention on it.

PR #72, opened 2020-05-24
Allow java-ea to fail

ameyjadiye:


ameyjadiye:
@garydgregory , please pull this, this will reflect correct build status for upcoming PRs.

PR #73, opened 2020-05-25
Refactor CSVFormat to use an internal builder

dota17:
refactor CSVFormat withXXX methods using a new constructor

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -726,6 +726,8 @@ public final class CSVFormat implements Serializable {
 
     private final boolean trim;
 
+    private final Builder builder;
+
     /**
      * Creates a customized CSV format.
      *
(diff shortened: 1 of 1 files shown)


garydgregory:
I'm not sure if replacing usage of a type-safe constructor with a type-unsafe `Map` (values are `Object`) is a good thing. What do others think? For future maintainers, it will be easy (IMO) to mess up the cast-typing and the key strings.

melloware:
I agree with @garydgregory it should be an object passed with concrete values not a Hashmap of keys. I think it needs to be a concrete POJO to pass it.

garydgregory:
@melloware 
The POJO is the `CSVFormat` itself... so there is nothing to do with this PR IMO.

melloware:
Agreed.

dota17:
@garydgregory @melloware 
I have completely reworked the code for refactorCSVFormat. This modification is mainly because the new CSVFormat()  will be called many times, and there are many parameters, such code doesn't look very goog,  and when adding an attribute more code is modified.
 It may be better to use the Builder mode, but the Builder mode will cause the API to change.

melloware:
@dota17 i will admit this does look a little nicer but i am not sure introducing reflection just for the sake of rearranging the code.

I do appreciate what you are trying to do though because the code does look "wordy" in its current state.

garydgregory:
-1 Reflection is a hack of last resort IMO, not something to use just for style.

dota17:
@garydgregory @melloware 
How about this way of modifying code?

melloware:
This does look a lot better.

garydgregory:
@dota17 and all:

I've brought this patch in locally and it does not make sense to me. 

The new `Builder` class does not, in fact, follow the builder pattern as it does not build anything, it does not have a `build()` method. That's easy to fix, so I did that (locally) and it still does not make sense. 

The point of a builder class is usually to call the methods on a single builder (in the fluent style or not, that does not matter) and then build a single domain object. So you have two instances in play. The builder which you can call 100 times, and then a single call to a `build()` method to give you a new domain object, here a `CSVFormat`. 

Without this patch, if you call 100 `CSVFormat#with` methods, you get 100 new `CSVFormat `objects, whether you need them or not, which is not great. With this PR, if you call 100 C`SVFormat#with` methods, not only do you still get 100 new `CSVFormat` objects, but you also get 100 new `Builder` objects! It makes the current situation in Git master even worse! For what? I don't get it.

I think what we really want is a single instance of a Builder object that you can call 100 times which creates no extra Builder objects and no extra domain objects. This PR does the opposite, so -1.

garydgregory:
Closing, see git master.

PR #77, opened 2020-06-17
Fix null, emptyString, nullString

dota17:
* https://issues.apache.org/jira/projects/CSV/issues/CSV-93
* https://issues.apache.org/jira/projects/CSV/issues/CSV-253

When QuoteMode: ALL_NON_NULL (Quotes all non-null fields), we can distinguish null through it whether it is quoted. 
So ```"NULL",NULL``` should be parsed ```["NULL", null]```, but now it is parsed ```[null, null]```, when ```withNullString("NULL")```. And ```"",``` should be parsed ```["", null]```, but it is parsed ```["",""]```.
QuoteMode: NON_NUMERIC same as ALL_NON_NULL.

This PR can solve this problem. if user want to distinguish empty string and absent value, they can set QuoteMode to All_NON_NULL or NON_NUMERIC.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVParser.java
+++ b/src/main/java/org/apache/commons/csv/CSVParser.java
@@ -380,6 +380,10 @@ public final class CSVParser implements Iterable<CSVRecord>, Closeable {
 
     private final Token reusableToken = new Token();
 
+    private final String nullString;
+
+    private final boolean strictQuoteMode;
+
     /**
      * Customized CSV parser using the given {@link CSVFormat}
      *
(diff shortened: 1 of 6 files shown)


lbruun:
If we go this route rather than [PR 51](https://github.com/apache/commons-csv/pull/51) then the Javadoc for `QuoteMode` need to be updated too. At the moment it clearly states that `QuoteMode` is only used when printing (creating CSV output), _not_ when parsing.

garydgregory:
Merged as commit ac280e705d99cc1168f93f7d9a5f92618bebbb97.

PR #114, opened 2020-10-02
CSV-264: Added DuplicateHeaderMode for flexibility with header strictness.

SethFalco:
Instead of only having a boolean `true`/`false` for how duplicate header values are handled, this uses an enum instead.

Previously the only possibilities were:
* `true`: To allow duplicates.
* `false`: To disallow duplicates, except empty cells which were allowed to be duplicates.

This pull request makes an enum with three options:
* `ALLOW_ALL`: To always allow duplicates. (Same as `true` previously.)
* `ALLOW_EMPTY`: To allow duplicates only if they're empty cells. (Same as `false` previously.)
* `DISALLOW`: To disallow all duplicates. 

This provides a little more flexibility in the strictness for the parser, and makes what the options do clearer.

Jira Issue: https://issues.apache.org/jira/browse/CSV-264

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -17,6 +17,18 @@
 
 package org.apache.commons.csv;
 
+import static org.apache.commons.csv.Constants.BACKSLASH;
+import static org.apache.commons.csv.Constants.COMMA;
+import static org.apache.commons.csv.Constants.COMMENT;
+import static org.apache.commons.csv.Constants.CR;
+import static org.apache.commons.csv.Constants.CRLF;
+import static org.apache.commons.csv.Constants.DOUBLE_QUOTE_CHAR;
+import static org.apache.commons.csv.Constants.EMPTY;
+import static org.apache.commons.csv.Constants.LF;
+import static org.apache.commons.csv.Constants.PIPE;
+import static org.apache.commons.csv.Constants.SP;
+import static org.apache.commons.csv.Constants.TAB;
+
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -35,18 +47,6 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
-
-import static org.apache.commons.csv.Constants.BACKSLASH;
-import static org.apache.commons.csv.Constants.COMMA;
-import static org.apache.commons.csv.Constants.COMMENT;
-import static org.apache.commons.csv.Constants.CR;
-import static org.apache.commons.csv.Constants.CRLF;
-import static org.apache.commons.csv.Constants.DOUBLE_QUOTE_CHAR;
-import static org.apache.commons.csv.Constants.EMPTY;
-import static org.apache.commons.csv.Constants.LF;
-import static org.apache.commons.csv.Constants.PIPE;
-import static org.apache.commons.csv.Constants.SP;
-import static org.apache.commons.csv.Constants.TAB;
 
 /**
  * Specifies the format of a CSV file and parses input.
@@ -1473,10 +1473,10 @@ public final class CSVFormat implements Serializable {
     }
 
     /**
-     * Returns how duplicate headers are handled.
+     * Gets how duplicate headers are handled.
      *
      * @return if duplicate header values are allowed, allowed conditionally, or disallowed.
-     * @since 1.9
+     * @since 1.9.0
      */
     public DuplicateHeaderMode getDuplicateHeaderMode() {
         return duplicateHeaderMode;
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -17,18 +17,6 @@
 
 package org.apache.commons.csv;
 
-import static org.apache.commons.csv.Constants.BACKSLASH;
-import static org.apache.commons.csv.Constants.COMMA;
-import static org.apache.commons.csv.Constants.COMMENT;
-import static org.apache.commons.csv.Constants.CR;
-import static org.apache.commons.csv.Constants.CRLF;
-import static org.apache.commons.csv.Constants.DOUBLE_QUOTE_CHAR;
-import static org.apache.commons.csv.Constants.EMPTY;
-import static org.apache.commons.csv.Constants.LF;
-import static org.apache.commons.csv.Constants.PIPE;
-import static org.apache.commons.csv.Constants.SP;
-import static org.apache.commons.csv.Constants.TAB;
-
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -47,6 +35,18 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
+
+import static org.apache.commons.csv.Constants.BACKSLASH;
+import static org.apache.commons.csv.Constants.COMMA;
+import static org.apache.commons.csv.Constants.COMMENT;
+import static org.apache.commons.csv.Constants.CR;
+import static org.apache.commons.csv.Constants.CRLF;
+import static org.apache.commons.csv.Constants.DOUBLE_QUOTE_CHAR;
+import static org.apache.commons.csv.Constants.EMPTY;
+import static org.apache.commons.csv.Constants.LF;
+import static org.apache.commons.csv.Constants.PIPE;
+import static org.apache.commons.csv.Constants.SP;
+import static org.apache.commons.csv.Constants.TAB;
 
 /**
  * Specifies the format of a CSV file and parses input.
@@ -188,8 +188,6 @@ public final class CSVFormat implements Serializable {
             return new Builder(csvFormat);
         }
 
-        private boolean allowDuplicateHeaderNames;
-
         private boolean allowMissingColumnNames;
 
         private boolean autoFlush;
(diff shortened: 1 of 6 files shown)


aherbert:
Thanks for the PR. You have removed some public API methods. This is not allowed and would fail binary compatibility tests. I have enabled these on the CI environments. If you rebase on master the errors should appear in the checks.

You can run:
```
mvn
```

on the project to see the full checks on the code. 

The `mvn clirr:check` goal will fail and indicate the public methods that have been removed. These should be supported in addition to the new methods with the new duplicate header names enum.

SethiPandi:
Sorry about that, I rebased master and applied the changes requested.
I believe this should be good now, though I'll take a peek at the unsuccessful checks now.

I usually use `git pull` and `git merge` to obtain changes, I'm a bit new to using `rebase` however.  
I'm not 100% sure if I was supposed to use `git push --force-with-lease...`, but the usual `git push...` was throwing warnings, please let me know if that's a problem.

garydgregory:
You can force push to your branch all you want; -) we usually do not rewrite history for long lived branches like master and release.

SethiPandi:
Not sure if you guys are interested in merging this PR, but figured I'd fix the merge conflicts that appeared from recent commits.

garydgregory:
@SethiPandi 
Please rebase on master.
TY!
Gary

SethiPandi:
Just a heads up that this is rebased with `master` now!

garydgregory:
@SethFalco 
Please accept my apologieses for the delay in further reviewing this PR.

I feel there is something important missing here:

- What happens/should happen for each enum value when you call `CSVRecord.get(String)`.
- What happens/should happen for each enum value when you call `CSVRecord.toMap()`.

For example, if duplicate headers are allowed, does the first column win? Should it? Does the last column win? Certainly, it should not be random or undocumented.

Also, would you mind rebasing on master please? I've recently deprecated the `with` methods in favor of a Builder. New settings should be added through the Builder, not `with` methods.

SethFalco:
It's worth noting this change doesn't impact how data is resolved internally or how headers are referenced, but just how if duplicate empty headers are allowed. I can peek into the current behavior anyway and try to document it if it's not already.

Meanwhile, I've rebased with master.
I also squashed my commits to make rebasing a bit easier.

I also updated an exception to no longer recommend using one of the deprecated methods.

SethFalco:
I didn't see it documented, figured the easiest way to find out would be to just test it.
It currently uses the last column with the given name.

> What happens/should happen for each enum value when you call CSVRecord.get(String).

| A | A
|---|---
| 1 | 2
| 3 | 4

```java
List<CSVRecord> records = parser.getRecords();

for (CSVRecord record : records)
    System.out.println(record.get("A"));

// 2
// 4
```

> What happens/should happen for each enum value when you call CSVRecord.toMap().

| A | A | B | B
|---|---|---|---
| 1 | 2 | 5 | 6
| 3 | 4 | 7 | 8

```java
List<CSVRecord> records = parser.getRecords();

for (CSVRecord record : records) {
    Map<String, String> map = record.toMap();
    System.out.println(map);
}

// {A=2, B=6}
// {A=4, B=8}
```

I think we could add a docstring on `CSVRecord#get` and `CSVRecord#toMap` to clarify this.
We could also add something on `CSVFormat#setDuplicateHeaderMode`?

It might be a good idea to add test cases to cover this scenario, if it'll become documented behavior.

Mind if I do this in a separate PR?
I'd say it counts as a separate change from this one, so it'll keep things tidier.

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -1439,9 +1455,21 @@ public String format(final Object... values) {
...
     public boolean getAllowDuplicateHeaderNames() {
-        return allowDuplicateHeaderNames;
+        return duplicateHeaderMode == DuplicateHeaderMode.ALLOW_ALL;
+    }
+
+    /**
+     * Returns how duplicate headers are handled.
+     *
+     * @return if duplicate header values are allowed, allowed conditionally, or disallowed.
+     * @since 1.9
```
garydgregory, line 1469: "1.9" -> "1.9.0"

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -1439,9 +1455,21 @@ public String format(final Object... values) {
...
+     * @deprecated Use {@link #getDuplicateHeaderMode()}.
      */
+    @Deprecated
     public boolean getAllowDuplicateHeaderNames() {
-        return allowDuplicateHeaderNames;
+        return duplicateHeaderMode == DuplicateHeaderMode.ALLOW_ALL;
+    }
+
+    /**
+     * Returns how duplicate headers are handled.
```
garydgregory, line 1466: "Returns" -> "Gets"

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -17,8 +17,6 @@
 
 package org.apache.commons.csv;
 
-import static org.apache.commons.csv.Token.Type.TOKEN;
```
garydgregory, line 20: Leave the order as is, please.

src/main/java/org/apache/commons/csv/DuplicateHeaderMode.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public enum DuplicateHeaderMode {
+
+    /**
+     * Allows all duplicate headings.
+     */
+    ALLOW_ALL,
+
+    /**
+     * Allows duplicate headings only if they're empty
```
garydgregory, line 34: "headings" -> "headers"

src/main/java/org/apache/commons/csv/DuplicateHeaderMode.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Determines how duplicate header fields should be handled
+ * if {@link CSVFormat#withHeader(String...)} is not null.
+ *
+ * @since 1.9
+ */
+public enum DuplicateHeaderMode {
+
+    /**
+     * Allows all duplicate headings.
```
garydgregory, line 29: "headings" -> "headers"

src/main/java/org/apache/commons/csv/DuplicateHeaderMode.java
```diff
@@ -0,0 +1,43 @@
...
+    ALLOW_ALL,
+
+    /**
+     * Allows duplicate headings only if they're empty
+     * strings or null.
+     */
+    ALLOW_EMPTY,
+
+    /**
+     * Disallows duplicate headings entirely.
```
garydgregory, line 40: "headings" -> "headers"

src/test/java/org/apache/commons/csv/issues/JiraCsv264Test.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.StringReader;
+
+/**
+ * When {@link CSVFormat#withHeader(String...)} is not null;
+ * duplicate headers with empty strings should not be allowed.
+ *
+ * @see <a href="https://issues.apache.org/jira/projects/CSV/issues/CSV-264?filter=allopenissues">Jira Ticker</a>
```
garydgregory, line 32: Simplify link, use https://issues.apache.org/jira/browse/CSV-264

src/test/java/org/apache/commons/csv/issues/JiraCsv264Test.java
```diff
@@ -0,0 +1,98 @@
...
+            .setHeader()
+            .setDuplicateHeaderMode(DuplicateHeaderMode.DISALLOW)
+            .setAllowMissingColumnNames(true)
+            .build();
+
+        try (StringReader reader = new StringReader(CSV_STRING)) {
+            try {
+                csvFormat.parse(reader);
+                Assertions.fail("This shouldn't parse the CSV string successfully.");
+            } catch (IllegalArgumentException ex) {
```
garydgregory, line 60: Please use the JUnit API `Assertions.assertThrows(...)`

src/test/java/org/apache/commons/csv/issues/JiraCsv264Test.java
```diff
@@ -0,0 +1,98 @@
...
+            .setHeader()
+            .setDuplicateHeaderMode(DuplicateHeaderMode.DISALLOW)
+            .setAllowMissingColumnNames(true)
+            .build();
+
+        try (StringReader reader = new StringReader(CSV_STRING_GAP)) {
+            try {
+                csvFormat.parse(reader);
+                Assertions.fail("This shouldn't parse the CSV string successfully.");
+            } catch (IllegalArgumentException ex) {
```
garydgregory, line 93: Please use the JUnit API `Assertions.assertThrows(...)`

garydgregory (changes requested):
@SethFalco 
Thank you for the update. Please see my comments scattered throughout.
Gary

SethFalco:
Sorry about the imports, I'm guessing that due to a setting in IntelliJ setting back when I was using that.
I've rebased with master and done the requested changes. (I hope!)

garydgregory:
@SethFalco 
TY, the PR looks good to me. I'd like feedback from the community as well.

SethFalco:
Just rebased this to resolve merge conflicts.

> I'd like feedback from the community as well.

Seems no one wants to comment on this. ^-^' 

Any chance this could be merged soon, or still waiting on feedback? (Note, I'm in no rush.)

kinow (approved):
It sounds like a good fix for the issue described in JIRA. Only adds one more setting, giving more flexibility for users.

+1 and thanks for the contribution and patience @SethFalco !

-Bruno

PR #127, opened 2020-12-18
Minor Improvements

arturobernalg:
* Add final
* Unnecessary semicolon ''
* Use StandardCharsets
* Fix javadoc

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -1402,7 +1402,6 @@ public final class CSVFormat implements Serializable {
         int start = 0;
         int pos = 0;
         final int len = value.length();
-        final int end = len;
 
         final char delimChar = getDelimiter();
         final char quoteChar = getQuoteCharacter().charValue();
@@ -1445,7 +1444,7 @@ public final class CSVFormat implements Serializable {
                     // by including the default comment char too.
                     quote = true;
                 } else {
-                    while (pos < end) {
+                    while (pos < len) {
                         c = value.charAt(pos);
                         if (c == LF || c == CR || c == quoteChar || c == delimChar || c == escapeChar) {
                             quote = true;
@@ -1455,7 +1454,7 @@ public final class CSVFormat implements Serializable {
                     }
 
                     if (!quote) {
-                        pos = end - 1;
+                        pos = len - 1;
                         c = value.charAt(pos);
                         // Some other chars at the end caused the parser to fail, so for now
                         // encapsulate if we end in anything less than ' '
@@ -1468,7 +1467,7 @@ public final class CSVFormat implements Serializable {
 
             if (!quote) {
                 // no encapsulation needed - write out the original value
-                out.append(value, start, end);
+                out.append(value, start, len);
                 return;
             }
             break;
@@ -1478,7 +1477,7 @@ public final class CSVFormat implements Serializable {
 
         if (!quote) {
             // no encapsulation needed - write out the original value
-            out.append(value, start, end);
+            out.append(value, start, len);
             return;
         }
 
@@ -1487,7 +1486,7 @@ public final class CSVFormat implements Serializable {
 
         // Pick up where we left off: pos should be positioned on the first character that caused
         // the need for encapsulation.
-        while (pos < end) {
+        while (pos < len) {
             final char c = value.charAt(pos);
             if (c == quoteChar || c == escapeChar) {
                 // write out the chunk up until this point
@@ -1507,6 +1506,7 @@ public final class CSVFormat implements Serializable {
      * Always use quotes unless QuoteMode is NONE, so we not have to look ahead.
      *
      * @throws IOException
+     *             If an I/O error occurs
      */
     private void printWithQuotes(final Reader reader, final Appendable out) throws IOException {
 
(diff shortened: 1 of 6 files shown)


src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -572,7 +572,7 @@ public void testGetOneLine() throws IOException {
     /**
      * Tests reusing a parser to process new string records one at a time as they are being discovered. See [CSV-110].
      *
-     * @throws IOException
+     * @throws Exception Any exception can be thrown
```
garydgregory, line 575: Does not match the code.

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -1629,6 +1629,7 @@ private CharSequence trim(final CharSequence charSequence) {
      * Verifies the consistency of the parameters and throws an IllegalArgumentException if necessary.
      *
      * @throws IllegalArgumentException
+     *             thrown if the delimiter is a scape character
```
garydgregory, line 1632: "scape"?

garydgregory (changes requested):

src/test/java/org/apache/commons/csv/CSVParserTest.java
arturobernalg: changed

src/main/java/org/apache/commons/csv/CSVFormat.java
arturobernalg: changed. What do you thing?=

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -572,7 +572,7 @@ public void testGetOneLine() throws IOException {
     /**
      * Tests reusing a parser to process new string records one at a time as they are being discovered. See [CSV-110].
      *
-     * @throws IOException
+     * @throws IOException If something goes wrong
```
garydgregory, line 575: Might as well use the "standard" comment here

src/test/java/org/apache/commons/csv/PerformanceTest.java
```diff
@@ -64,7 +64,7 @@
     private static int max = 11; // skip first test
 
     private static int num = 0; // number of elapsed times recorded
-    private static long[] elapsedTimes = new long[max];
+    private static final long[] elapsedTimes = new long[max];
```
garydgregory, line 67: Constants are usually in uppercase.

src/test/java/org/apache/commons/csv/CSVFileParserTest.java
```diff
@@ -135,7 +136,7 @@ public void testCSVUrl(final File testFile) throws Exception {
 
             // Now parse the file and compare against the expected results
             final URL resource = ClassLoader.getSystemResource("org/apache/commons/csv/CSVFileParser/" + split[0]);
-            try (final CSVParser parser = CSVParser.parse(resource, Charset.forName("UTF-8"), format)) {
+            try (final CSVParser parser = CSVParser.parse(resource, StandardCharsets.UTF_8, format)) {
```
garydgregory, line 139: Good one :-)

src/test/java/org/apache/commons/csv/CSVParserTest.java
arturobernalg: Sorry Garry, but What is the standard comment?

src/test/java/org/apache/commons/csv/PerformanceTest.java
arturobernalg: true. changed

src/test/java/org/apache/commons/csv/CSVParserTest.java
garydgregory: @arturobernalg 
The JRE uses "if an I/O error occurs." but I prefer "Thrown when an I/O error occurs."

arturobernalg: ok, got it. Change it.

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -1628,7 +1628,8 @@ private CharSequence trim(final CharSequence charSequence) {
     /**
      * Verifies the consistency of the parameters and throws an IllegalArgumentException if necessary.
      *
-     * @throws IllegalArgumentException
+     * @throws IllegalArgumentException If something goes wrong
```
garydgregory, line 1631: Hi @arturobernalg 
This change does not help anyone, and it's confusing when the line above tells you when the exception would be thrown. I've updated git master for this Javadoc. Please rebase on master. TY.

arturobernalg: OK. Got it. Changed

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -427,7 +427,7 @@ public void testExcelFormat2() throws Exception {
     /**
      * Tests an exported Excel worksheet with a header row and rows that have more columns than the headers
      *
-     * @throws Exception
+     * @throws Exception Any exception can be thrown
```
garydgregory, line 430: Not helpful. Tests like this one can throw `Exception` as a convention instead of throwing a potentially long list of exceptions which are not useful and just clutter up the method signatures. So either we should not have a Javadoc tag or if we have one, it should carry proper and useful documentation. In this case, I think that there is nothing interesting to say.

arturobernalg: OK, got it. Reverted

PR #160, opened 2021-07-08
Update gitignore to ignore idea and vscode

SethFalco:
I'm about to rebase and respond to your feedback on https://github.com/apache/commons-csv/pull/114.
Just wanted to propose some minor changes, though.

### .gitignore
* Ignores all IntelliJ specific files. (What I was using before, so had some files left around still.)
* Ignores all Visual Studio Code specific files. (What I'm using now.)
* Replaces the dead link at the top of the file.

Edit: While I was at it, I opted to update the contents of .gitignore with the up-to-date revision on [github/gitignore](https://github.com/github/gitignore).

I can replace the link with `https://raw.githubusercontent.com/github/gitignore/6651bca9c83572f908474fc6206ed8a80df91290/Maven.gitignore` if preferred?
I prefer the current link, but whatever suits you.

### ~~.editorconfig~~
~~For those using code editors like Visual Studio Code, it can be easier if the project defines its preferences using [EditorConfig](https://editorconfig.org/) to override any program settings. (a vendor-agnostic configuration file that many code editors support)~~

~~I've just taken the EditorConfig configuration from the most starred Java project Apache has on GitHub:
https://github.com/apache/dubbo/blob/3.0/.editorconfig~~

~~It could be worth considering if you'd want to change anything about it, for example I think something like this might be better:~~
```
root = true

[*]
indent_style = space
indent_size = 2
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true

# Preserve line breaks in Markdown
[*.md]
trim_trailing_whitespace = false

# 4 space indentation
[*.{java,xml}]
indent_size = 4
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-### https://raw.github.com/github/gitignore/14b7566ce157ce95b07006466bacee160f242284/maven.gitignore
+### https://github.com/github/gitignore/blob/master/Maven.gitignore
 
 target/
 pom.xml.tag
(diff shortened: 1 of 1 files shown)


garydgregory:
We don't use editorconfig anywhere in Commons that I'm aware of so let's skip it. If anything , we could store formatting configs for Eclipse, IJ, and so on in src/conf, maybe...

SethFalco:
Just made one more minor edit, sorry!
About to look at my PRs to commons-beanutils again since I left those hanging. ^-^'

Realized `.settings` can be in, submodule directories as well. Thought I should update that too.

PR #163, opened 2021-07-13
CSV-279: Optimize Lexer Delimiter Check for One Character Delimiter

belugabehr:
If the delimiter is multiple characters, some amount of processing needs to happen to support it.  However, if the delimiter is a single character (common scenario), then the additional multi-byte processing (which are no-ops) can be skipped.

https://issues.apache.org/jira/browse/CSV-279


garydgregory:
This PR has no description or URL to the Jira ticket.

belugabehr:
@garydgregory Please let me know if I have updated the PR with the required info.  Thanks.

PR #168, opened 2021-07-15
CSV-284: Formalize PerformanceTest

belugabehr:
> There is a class org.apache.commons.csv.perf.PerformanceTest which is not documented and does not work without some hand-edits.
> Re-enable this test, clean it up, and document it.

https://issues.apache.org/jira/browse/CSV-284

### DIFF
--- a/src/test/java/org/apache/commons/csv/perf/PerformanceTest.java
+++ b/src/test/java/org/apache/commons/csv/perf/PerformanceTest.java
@@ -19,7 +19,6 @@ package org.apache.commons.csv.perf;
 
 import java.io.BufferedReader;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/csv/perf/PerformanceTest.java
```diff
@@ -85,24 +85,12 @@ private long parse(final Reader reader, final boolean traverseColumns) throws IO
         return recordCount;
     }
 
-    private void println(final String s) {
-        System.out.println(s);
-    }
-
-    private long readAll(final BufferedReader in) throws IOException {
```
garydgregory, line 92: I disagree here, keep this as is so it can be commented out or changed easily when testing.

belugabehr: You got it.

PR #191, opened 2021-10-25
CSV-292: Add Automatic-Module-Name to JAR file

rvesse:
Ensure that the resulting Commons CSV JAR file can be used in JPMS based
projects

Prior to this change the JAR is missing the `Automatic-Module-Name` entry in its
manifest meaning it cannot be used in JPMS based projects (e.g. rvesse/airline#106) yielding errors like the following:

```
Error occurred during initialization of boot layer
java.lang.module.FindException: Module org.apache.commons.csv not found, required by com.github.rvesse.airline.examples
```

With this change JPMS based projects can successfully use Commons CSV without errors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -285,6 +285,24 @@
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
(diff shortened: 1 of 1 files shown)


rvesse:
@kinow What needs to happen for this to get merged?

kinow:
> @kinow What needs to happen for this to get merged?

Ops, missed this PR, sorry @rvesse ! LGTM

PR #249, opened 2022-08-09
[CSV-300] Return new ArrayList in CSVRecord.toList() to prevent access to private values array

sman-81:
This PR offers a fix to bug CSV-300.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVRecord.java
+++ b/src/main/java/org/apache/commons/csv/CSVRecord.java
@@ -19,6 +19,7 @@ package org.apache.commons.csv;
 
 import java.io.Serializable;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -248,7 +249,8 @@ public final class CSVRecord implements Serializable, Iterable<String> {
     }
 
     /**
-     * Returns an iterator over the values of this record.
+     * Returns an iterator over the values of this record.<br>
+     * The iterator is read-only i.e. {@link Iterator#remove()} is not supported.
      *
      * @return an iterator over the values of this record.
      */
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/csv/CSVRecord.java
```diff
@@ -298,13 +299,14 @@ public Stream<String> stream() {
...
+     * Converts the values to a new {@code List}.<br>
+     * Modifications to the list do not alter this record.
      *
-     * @return a new List
+     * @return a new {@code List}
      * @since 1.9.0
      */
     public List<String> toList() {
-        return Arrays.asList(values);
+        return new ArrayList<>(Arrays.asList(values));
```
kinow, line 309: See @aherbert 's comment in the JIRA linked about returning an immutable list. Unless there's a reason for allowing users to modify the returned list, I think an immutable list would be best :+1: 

Thanks for the pull request!
-Bruno

garydgregory:
-1 needs a unit test

src/main/java/org/apache/commons/csv/CSVRecord.java
kinow: (Also needs to update the javadocs as mentioned in the JIRA comment too, and needs unit tests as pointed below by @garydgregory :+1: )

garydgregory: I don't see the point in returning a half-functional list. The list should not mangle the underlying record that's all.

sman-81:
Thanks for your feedback. PR updated.

garydgregory:
https://issues.apache.org/jira/browse/CSV-300

garydgregory:
Hm, I don't see the point of a read only list. IMO if I convert the record to a list, I should be able to do whatever I want with the list. If the list is RO, then I have to make a copy to edit it which is counter intuitive and wasteful. This is a toList() convert method and not a getList() access method after all. The point of a list here IMO is that I can edit its size, unlike an array.

sman-81:
Reason for closing pr: @garydgregory implemented his own solution in master, making my pr obsolete.

PR #250, opened 2022-08-11
[CSV-301] Fix warnings (mostly deprecation) in unit tests

sman-81:
This PR updates unit tests to use the library's new API and thus remove all warnings (mostly deprecation) in tests.

### DIFF
--- a/src/test/java/org/apache/commons/csv/CSVFileParserTest.java
+++ b/src/test/java/org/apache/commons/csv/CSVFileParserTest.java
@@ -64,17 +64,17 @@ public class CSVFileParserTest {
             final String[] split = line.split(" ");
             assertTrue(split.length >= 1, testFile.getName() + " require 1 param");
             // first line starts with csv data file name
-            CSVFormat format = CSVFormat.newFormat(',').withQuote('"');
+            CSVFormat format = CSVFormat.newFormat(',').builder().setQuote('"').build();
             boolean checkComments = false;
             for (int i = 1; i < split.length; i++) {
                 final String option = split[i];
                 final String[] option_parts = option.split("=", 2);
                 if ("IgnoreEmpty".equalsIgnoreCase(option_parts[0])) {
-                    format = format.withIgnoreEmptyLines(Boolean.parseBoolean(option_parts[1]));
+                    format = format.builder().setIgnoreEmptyLines(Boolean.parseBoolean(option_parts[1])).build();
                 } else if ("IgnoreSpaces".equalsIgnoreCase(option_parts[0])) {
-                    format = format.withIgnoreSurroundingSpaces(Boolean.parseBoolean(option_parts[1]));
+                    format = format.builder().setIgnoreSurroundingSpaces(Boolean.parseBoolean(option_parts[1])).build();
                 } else if ("CommentStart".equalsIgnoreCase(option_parts[0])) {
-                    format = format.withCommentMarker(option_parts[1].charAt(0));
+                    format = format.builder().setCommentMarker(option_parts[1].charAt(0)).build();
                 } else if ("CheckComments".equalsIgnoreCase(option_parts[0])) {
                     checkComments = true;
                 } else {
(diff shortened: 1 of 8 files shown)


garydgregory:
-1 for now, because, we MUST test all existing Commons CSV APIs. I do not think we want to write _new_ tests that only test deprecated APIs.

sman-81:
@garydgregory 
The Builder api was added over a year ago (July 2021) and the CSVFormat#with methods deprecated.
~300 deprecation warnings in master are super-distracting for anybody contributing to this project.
Time to release 2.0.0?

> -1 for now, because, we MUST test all existing Commons CSV APIs.
deprecated APIs including their implementation should not be changed.
If their tests have been green, tests can be updated as is done in my pr.

garydgregory:
> Isn't it about time to release 2.0.0?

That's a different issue from this PR and more appropriate for the developer's mailing list.

No one has asked for a major release that I am aware of. Usually, we release a major release to break the API for new features. A major release to only drop deprecated methods seems unnecessary (to me) for now. 

My main concern is making sure the deprecated API remains the same and is thoroughly tested. Maybe the code coverage action on GitHub is enough, I am not sure, but it might be a good sign that it has not complained.

Let's see what others think.

sman-81:
What I was trying to say, release 2.0.0 _with deprecated API removed_ and _all code that uses deprecated API updated_. This PR is obviously just part of that work.
Noone is talking about a major release merely to update some unit tests, that would be complete nonsense.
Having more frequent releases (a couple per year) would be desirable for this project as it would be for most libraries imo.

garydgregory:
@kinow Any thoughts?

kinow:
> Having more frequent releases (a couple per year) would be desirable for this project as it would be for most libraries imo.

Agreed! Unfortunately we are a little restricted on developer-bandwidth. But if you are willing to volunteer pushing some work for 2.0.0, that'd be really helpful!

>-1 for now, because, we MUST test all existing Commons CSV APIs. I do not think we want to write new tests that only test deprecated APIs.

@sman-81 I think a 2.0.0 release would be great, but we probably would have to start reviewing issues in JIRA, existing pull requests, and the status of the code base (e.g. do we need to bump java version for any good reason [we normally avoid unnecessary minimum JVM bumps], is there anything useful in existing CSV libraries that would be worth discussing adding in 2.0.0, etc.)

I agree we must not remove tests based on deprecation warnings, as they can still be useful. I also agree that deprecation warnings can be demotivating to contributors, and even to long-time maintainers/developers. I use Eclipse & IntelliJ, and I always crave to have as little warnings as I can.

But I think in Eclipse & IntelliJ we should be able to suppress the warnings for now, while we work on a 2.0.0 release. WDYT @sman-81 ?

We do not need to close this PR, if you'd like. Maybe make it a draft?

@garydgregory I think you have a lot more experience with the CSV code base & maintaining the component. Do you think, if we fix existing issues in JIRA, review the PR's, and remove deprecated API, we could prepare a 2.0.0 release? I know we have a longer time to release major releases, but one or two major releases per year doesn't sound too bad IMO. WDYT?

garydgregory:
There is value in a new release of course, I don't see the point of a major release since we do not have a feature driver to break binary compatibility. Don't forget that a major release entails a package name and artifact name change, and there is no reason for that looking at the current code base IMO.

kinow:
Good point @garydgregory  , I just logged in to JIRA to check the CSV project releases, and looks like the 2.0 has no issues so far.

image

@sman-81 I guess CSV-301 could be the first added to 2.0.0. We can monitor new code, making sure it uses the new API and don't introduce new warnings. But not much else we can do until 2.0.0 I reckon.

sman-81:
@kinow, @garydgregory

Thanks for your good thoughts.

> Don't forget that a major release entails a package name and artifact name change, [..]

```
org.apache.commons:commons-csv2 package org.apache.commons.csv2
org.apache.commons:commons-csv3 package org.apache.commons.csv3
```

omg ... are you forced to do in this way by some set of conventions?

I personally prefer semver and that changes to the major version are entitled to break binary compatibility.

I don't know of anyone personally who replaces jars in a system without recompiling. I understand there is value in being able to do so (e.g. security patches). Does commons-csv or other Apache libraries publish and advertise this fact? It's kind of a unique feature, that comes at a price. If users don't know about the guaranteed binary compatibility, they would always recompile. I personally always recompile (working mostly on internal systems), as I would not want to take the risk of crashing with NoSuchMethodError at at run-time.

The deprecation warnings will be addressed while removing old API and old tests in a new major release.
We can close my PR - no problem at all. The work is easily recreated when releasing.

Again, thanks for your thoughts and feedback - have a great week.

kinow:
>omg ... are you forced to do in this way by some set of conventions?

@garydgregory might have a better explanation. I remember seeing that before I joined ASF. I also remember reading about it somewhere. I thought it was the Commons Wiki, but I couldn't find, I think:

- https://cwiki.apache.org/confluence/display/COMMONS/MavenGroupIDChange
- https://cwiki.apache.org/confluence/display/COMMONS/MavenAndClasspath

Maybe we should make this information more visible to contributors.

>The deprecation warnings will be addressed while removing old API and old tests in a new major release.
>We can close my PR - no problem at all. The work is easily recreated when releasing.

Thanks for understanding it! Closing the issue.

> Again, thanks for your thoughts and feedback - have a great week.

Thanks again and have a great week too @sman-81 !

garydgregory:
Hi all, @sman-81 , @kinow ,

Please see https://garygregory.wordpress.com/2020/06/14/how-we-handle-binary-compatibility-at-apache-commons/

sman-81:
> Please see https://garygregory.wordpress.com/2020/06/14/how-we-handle-binary-compatibility-at-apache-commons/

Interesting read. How about linking at your blog post from `CONTRIBUTING.md`?

We should also set up [revapi](https://revapi.org/revapi-site/main/index.html) to enforce binary compatibility i.e. to fail the build when non-compatible modifications are made to the public API.

sman-81:
> Please see https://garygregory.wordpress.com/2020/06/14/how-we-handle-binary-compatibility-at-apache-commons/

Interesting read. How about linking at your blog post from `CONTRIBUTING.md`?

We should also set up [revapi](https://revapi.org/revapi-site/main/index.html) or japicmp which is already mentioned to enforce binary compatibility i.e. to fail the build when non-compatible modifications are made to the public API.

PR #257, opened 2022-09-06
[CSV-304] Accessors for header/trailer comments

pedro-w:
Add accessors for header comments (before the header row)
and trailer comments (after the last record)
Also add javadoc and tests

See https://issues.apache.org/jira/browse/CSV-304

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVParser.java
+++ b/src/main/java/org/apache/commons/csv/CSVParser.java
@@ -352,6 +352,7 @@ public final class CSVParser implements Iterable<CSVRecord>, Closeable {
 
         return new CSVParser(new InputStreamReader(url.openStream(), charset), format);
     }
+
     private String headerComment;
 
     private String trailerComment;
@@ -487,8 +488,10 @@ public final class CSVParser implements Iterable<CSVRecord>, Closeable {
                 }
             } else {
                 if (this.format.getSkipHeaderRecord()) {
-                    final CSVRecord csvRecord = this.nextRecord();
-                    headerComment = csvRecord.getComment();
+                    final CSVRecord nextRecord = this.nextRecord();
+                    if (nextRecord != null) {
+                        headerComment = nextRecord.getComment();
+                    }
                 }
                 headerRecord = formatHeader;
             }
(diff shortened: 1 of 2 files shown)


kinow (commented):
One comment, but the rest of the code looks good! Thanks @pedro-w!

pedro-w:
I noticed that github is parsing the Javadoc annotation `@since` as a username. I apologize to the owner of that account.

kinow (approved):
Looks good to me. Asked a review from @garydgregory as he works on and knows a lot more about this component. Thanks @pedro-w !

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -1375,4 +1375,147 @@ private void validateRecordPosition(final String lineSeparator) throws IOExcepti
...
+    }
+    @Test
+    public void testGetTrailerComment_HeaderComment3() throws IOException {
+        try (CSVParser parser = CSVParser.parse(CSV_INPUT_HEADER_COMMENT, FORMAT_EXPLICIT_HEADER_NOSKIP)) {
+            parser.getRecords();
+            assertFalse(parser.hasTrailerComment());
+            assertNull(parser.getTrailerComment());
+        }
+    }
+    @Test
```
garydgregory, line 1512: Please separate methods with a blank line.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -713,6 +764,10 @@ CSVRecord nextRecord() throws IOException {
             case EOF:
                 if (this.reusableToken.isReady) {
                     this.addRecordValue(true);
+                } else {
```
garydgregory, line 767: Extra block level is not needed, IOW use "else if" instead of "else { if ... }"

garydgregory (changes requested):
Hi @pedro-w,
Is the intent to allow comments at the start and end of a file or before and after each record?
Please see my scattered comments as well.

pedro-w:
> 


> Is the intent to allow comments at the start and end of a file or before and after each record?

The former. I'll explain my thinking, *please correct me if I've got it wrong!*

    ; Comment 0
    A,B
    ; Comment 1
    1,1
    ; Comment 2
    2,2
    ; Comment 3

The comments attach to the following line so Comments 1 & 2 can currently be extracted from the corresponding `CSVRecord`s and Comment 3 can never be accessed. Comment 0 is a special case. If we ask `CSVParser` to generate the column headings from the first record, it can't be accessed, but otherwise it's attached to "A,B" which is just a normal data record.

My intention was to allow access to Comment 0 via `getHeaderComment` (only in the case of generated column names) and Comment 3 via `getTrailerComment` (always).


I'll get onto the other issues too as they're mostly just formatting.

PR #265, opened 2022-09-25
CSV-290 - Fix the wrong assumptions in PostgreSQL formats

angusdev:
I tested in psql 14.5 Homebrew in Mac M1.

CSVFormat.POSTGRESQL_CSV - special characters are not escaped.
CSVFormat.POSTGRESQL_TEXT - values are not quoted.

```sql
drop table COMMONS_CSV_PSQL_TEST;
create table COMMONS_CSV_PSQL_TEST (ID INTEGER, COL1 VARCHAR, COL2 VARCHAR, COL3 VARCHAR, COL4 VARCHAR);
insert into COMMONS_CSV_PSQL_TEST select 1, 'abc', 'test line 1' || chr(10) || 'test line 2', null, '';
insert into COMMONS_CSV_PSQL_TEST select 2, 'xyz', '\b:' || chr(8) || ' \n:' || chr(10) || ' \r:' || chr(13), 'a', 'b';
insert into COMMONS_CSV_PSQL_TEST values (3, 'a', 'b,c,d', '"quoted"', 'e');
copy COMMONS_CSV_PSQL_TEST TO '/tmp/psql.csv' WITH (FORMAT CSV);
copy COMMONS_CSV_PSQL_TEST TO '/tmp/psql.tsv';
```

```
cat /tmp/psql.csv
1,abc,"test line 1
test line 2",,""
2,xyz,"\b:^H \n:
\r:^M",a,b
3,a,"b,c,d","""quoted""",e
```

```
cat /tmp/psql.tsv
1    abc    test line 1\ntest line 2               \N
2    xyz    \\b:\b \\n:\n \\r:\r       a           b
3    a      b,c,d                      "quoted"    e
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -260,6 +260,8 @@
               <exclude>src/test/resources/org/apache/commons/csv/CSV-259/sample.txt</exclude>
               <exclude>src/test/resources/org/apache/commons/csv/CSVFileParser/testCSV246.csv</exclude>
               <exclude>src/test/resources/org/apache/commons/csv/CSVFileParser/testCSV246_checkWithNoComment.txt</exclude>
+              <exclude>src/test/resources/org/apache/commons/csv/CSV-290/psql.csv</exclude>
+              <exclude>src/test/resources/org/apache/commons/csv/CSV-290/psql.tsv</exclude>
             </excludes>
           </configuration>
         </plugin>
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -1047,7 +1047,7 @@ public final class CSVFormat implements Serializable {
      *
      * <p>
      * This is a comma-delimited format with a LF character as the line separator. Values are double quoted and special
-     * characters are escaped with {@code '"'}. The default NULL string is {@code ""}.
+     * characters are not escaped. The default NULL string is {@code ""}.
      * </p>
      *
      * <p>
@@ -1055,7 +1055,7 @@ public final class CSVFormat implements Serializable {
      * </p>
      * <ul>
      * <li>{@code setDelimiter(',')}</li>
-     * <li>{@code setEscape('"')}</li>
+     * <li>{@code setEscape(null)}</li>
      * <li>{@code setIgnoreEmptyLines(false)}</li>
      * <li>{@code setQuote('"')}</li>
      * <li>{@code setRecordSeparator('\n')}</li>
@@ -1071,7 +1071,7 @@ public final class CSVFormat implements Serializable {
     // @formatter:off
     public static final CSVFormat POSTGRESQL_CSV = DEFAULT.builder()
             .setDelimiter(COMMA)
-            .setEscape(DOUBLE_QUOTE_CHAR)
+            .setEscape(null)
             .setIgnoreEmptyLines(false)
             .setQuote(DOUBLE_QUOTE_CHAR)
             .setRecordSeparator(LF)
@@ -1084,8 +1084,8 @@ public final class CSVFormat implements Serializable {
      * Default PostgreSQL text format used by the {@code COPY} operation.
      *
      * <p>
-     * This is a tab-delimited format with a LF character as the line separator. Values are double quoted and special
-     * characters are escaped with {@code '"'}. The default NULL string is {@code "\\N"}.
+     * This is a tab-delimited format with a LF character as the line separator. Values are not quoted and special
+     * characters are escaped with {@code '\\'}. The default NULL string is {@code "\\N"}.
      * </p>
      *
      * <p>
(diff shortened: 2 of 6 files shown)


garydgregory:
Hi @angusdev 
Thank you for the PR.
Does the version of PostgreSQL matter?

angusdev:
I can't test previous versions of PostgreSQL. But it is pretty safe to say that it applies to all versions.

Postgresql support export to CVS since version 8.0 (year 2005, https://www.postgresql.org/docs/8.0/release-8-0.html)
In the documentation of 8.1 (https://www.postgresql.org/docs/8.1/sql-copy.html), it didn't state clearly but implied that the CSV export will not escape special characters.  The backslash escape is used for import data from text file (COPY FROM) only.

For text format (tab delimited), there is no reason to quote the text.

angusdev:
Please hold this PR.

Seems need to escape when writing , but no escape when reading PostgreSQL produced CSV. May need to create a separate CSVFormat (e.g. POSTGRESQL_CSV_READ), need sometime to test the behaviour.

angusdev:
Tested the behaviour of import and export are consistent.

Test case: export csv/tsv from PostgreSQL, read by commons-cvs and write to new csv/tsv, import to PostgreSQL, export csv/tsv again. compare the 1st and 2nd export file

```sql
drop table COMMONS_CSV_PSQL_TEST;
create table COMMONS_CSV_PSQL_TEST (ID INTEGER, COL1 VARCHAR, COL2 VARCHAR, COL3 VARCHAR, COL4 VARCHAR);
insert into COMMONS_CSV_PSQL_TEST select 1, 'abc', 'test line 1' || chr(10) || 'test line 2', null, '';
insert into COMMONS_CSV_PSQL_TEST select 2, 'xyz', '\b:' || chr(8) || ' \n:' || chr(10) || ' \r:' || chr(13), 'a', 'b';
insert into COMMONS_CSV_PSQL_TEST values (3, 'a', 'b,c,d', '"quoted"', 'e');
copy COMMONS_CSV_PSQL_TEST to '/tmp/psql.csv' with (FORMAT CSV);
copy COMMONS_CSV_PSQL_TEST to '/tmp/psql.tsv';
```

use commons-csv to read '/tmp/psql.csv' and write to '/tmp/outpsql.csv', same for 'psql.tsv'

```sql
truncate table COMMONS_CSV_PSQL_TEST;
copy COMMONS_CSV_PSQL_TEST(ID, COL1, COL2, COL3, COL4) from '/tmp/outpsql.csv' with (FORMAT CSV);
copy COMMONS_CSV_PSQL_TEST to '/tmp/psql2.csv' with (FORMAT CSV);

truncate table COMMONS_CSV_PSQL_TEST;
copy COMMONS_CSV_PSQL_TEST(ID, COL1, COL2, COL3, COL4) from '/tmp/outpsql.tsv';
copy COMMONS_CSV_PSQL_TEST to '/tmp/psql2.tsv';
```

diff /tmp/psql.csv /tmp/psql2.csv
(no difference)

diff /tmp/psql.tsv /tmp/psql2.tsv
(no difference)

garydgregory:
Hello @angusdev 

Thank you for updating your PR.

(1) I think you need to test for tab characters (ASCII 9) in values.

(2) In the PG docs I read
```
QUOTE
Specifies the quoting character to be used when a data value is quoted. The default is double-quote. This must be a single one-byte character. This option is allowed only when using CSV format.

ESCAPE
Specifies the character that should appear before a data character that matches the QUOTE value. The default is the same as the QUOTE value (so that the quoting character is doubled if it appears in the data). This must be a single one-byte character. This option is allowed only when using CSV format.
```
Please help me understand why the git master code does not match this definition.
TY!

angusdev:
Added test for tab characters (ASCII 9) in values.

For QUOTE and ESCAPE, see below example

```
postgres=# insert into COMMONS_CSV_PSQL_TEST select 1, '"quoted"', '|quoted2|', null, null;
INSERT 0 1
postgres=# copy COMMONS_CSV_PSQL_TEST to STDOUT with CSV;
1,"""quoted""",|quoted2|,,
postgres=# copy COMMONS_CSV_PSQL_TEST to STDOUT with CSV QUOTE '|';;
1,"quoted",|||quoted2|||,,
postgres=# copy COMMONS_CSV_PSQL_TEST to STDOUT with CSV ESCAPE '~';
1,"~"quoted~"",|quoted2|,,
postgres=# copy COMMONS_CSV_PSQL_TEST to STDOUT with CSV QUOTE '|' ESCAPE '~';
1,"quoted",|~|quoted2~||,,
postgres=# copy COMMONS_CSV_PSQL_TEST to STDOUT QUOTE '|';
ERROR:  COPY quote available only in CSV mode
postgres=# copy COMMONS_CSV_PSQL_TEST to STDOUT ESCAPE '~';
ERROR:  COPY escape available only in CSV mode
```

In PG (CSV), ESCAPE is used to escape the quote char, while in COMMONS_CSV, ESCAPE is to escape delimiter and special char

In PG (TEXT), QUOTE is not needed as it is tab-delimited and the delimiter (tab) is escaped by '\t'

PR #276, opened 2022-10-17
[CSV-302] CSVFormat.duplicateHeaderMode requires default DISALLOW

spannm:
- CSVFormat.duplicateHeaderMode requires default DISALLOW for backward compatibility
- The field does not allow null values
- Deserialization of version 1.9.0 CSVFormat objects now set duplicateHeaderMode based on the previous member boolean duplicateHeaderMode (done in readResolve)
- Add JUnit test for CSVFormat deserialization
- Fix a small bug in setNullString where member quotedNullString was inconsistently written (missing write in setQuote)

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -229,8 +229,6 @@ public final class CSVFormat implements Serializable {
 
         private Character quoteCharacter;
 
-        private String quotedNullString;
-
         private QuoteMode quoteMode;
 
         private String recordSeparator;
@@ -259,7 +257,6 @@ public final class CSVFormat implements Serializable {
             this.trailingDelimiter = csvFormat.trailingDelimiter;
             this.trim = csvFormat.trim;
             this.autoFlush = csvFormat.autoFlush;
-            this.quotedNullString = csvFormat.quotedNullString;
             this.duplicateHeaderMode = csvFormat.duplicateHeaderMode;
         }
 
@@ -370,7 +367,7 @@ public final class CSVFormat implements Serializable {
         /**
          * Sets the duplicate header names behavior.
          *
-         * @param duplicateHeaderMode the duplicate header names behavior
+         * @param duplicateHeaderMode the duplicate header names behavior, may not be null
          * @return This instance.
          * @since 1.10.0
          */
@@ -595,7 +592,6 @@ public final class CSVFormat implements Serializable {
          */
         public Builder setNullString(final String nullString) {
             this.nullString = nullString;
-            this.quotedNullString = quoteCharacter + nullString + quoteCharacter;
             return this;
         }
 
@@ -1338,26 +1334,24 @@ public final class CSVFormat implements Serializable {
     private final boolean trim;
 
     private CSVFormat(final Builder builder) {
-        this.delimiter = builder.delimiter;
-        this.quoteCharacter = builder.quoteCharacter;
-        this.quoteMode = builder.quoteMode;
-        this.commentMarker = builder.commentMarker;
-        this.escapeCharacter = builder.escapeCharacter;
-        this.ignoreSurroundingSpaces = builder.ignoreSurroundingSpaces;
-        this.allowMissingColumnNames = builder.allowMissingColumnNames;
-        this.ignoreEmptyLines = builder.ignoreEmptyLines;
-        this.recordSeparator = builder.recordSeparator;
-        this.nullString = builder.nullString;
-        this.headerComments = builder.headerComments;
-        this.header = builder.headers;
-        this.skipHeaderRecord = builder.skipHeaderRecord;
-        this.ignoreHeaderCase = builder.ignoreHeaderCase;
-        this.trailingDelimiter = builder.trailingDelimiter;
-        this.trim = builder.trim;
-        this.autoFlush = builder.autoFlush;
-        this.quotedNullString = builder.quotedNullString;
-        this.duplicateHeaderMode = builder.duplicateHeaderMode;
-        validate();
+        this(builder.delimiter,
+             builder.quoteCharacter,
+             builder.quoteMode,
+             builder.commentMarker,
+             builder.escapeCharacter,
+             builder.ignoreSurroundingSpaces,
+             builder.ignoreEmptyLines,
+             builder.recordSeparator,
+             builder.nullString,
+             builder.headerComments,
+             builder.headers,
+             builder.skipHeaderRecord,
+             builder.allowMissingColumnNames,
+             builder.ignoreHeaderCase,
+             builder.trim,
+             builder.trailingDelimiter,
+             builder.autoFlush,
+             builder.duplicateHeaderMode);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -2280,6 +2279,27 @@ private void validate() throws IllegalArgumentException {
...
+        // check whether field duplicateHeaderMode (as of 1.10.0) was found in serialized form
+        // if not, set from boolean allowDuplicateHeaderNames
+        if (duplicateHeaderMode == null) {
+            setFieldValue("duplicateHeaderMode",
+                    allowDuplicateHeaderNames ? DuplicateHeaderMode.ALLOW_ALL : DuplicateHeaderMode.DISALLOW);
+        }
+        return this;
+    }
+
+    private void setFieldValue(String fieldName, Object value) {
```
garydgregory, line 2292: Use `final` where you can.

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -2280,6 +2279,27 @@ private void validate() throws IllegalArgumentException {
...
+        try {
+            Field fld = getClass().getDeclaredField(fieldName);
+            fld.setAccessible(true); // make field non-final
+            fld.set(this, value);
+            fld.setAccessible(false);
+        } catch (Exception ex) {
+            throw new RuntimeException("Failed to set value of field " + fieldName, ex);
```
garydgregory, line 2299: RuntimeException is an anti-pattern, using IAE or ISE or another more descriptive exception.

src/test/java/org/apache/commons/csv/CSVFormatSerializationTest.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.ObjectInputStream;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+/**
+ * Test deserialization of class {@link CSVFormat}.
+ * @author Markus Spann
```
garydgregory, line 32: We don't use `author` tags. Credit is given in the changes.xml file post-merge.

garydgregory (changes requested):
Hi @sman-81 
Please see a few scattered comments.

src/main/java/org/apache/commons/csv/CSVFormat.java
sman-81: In Java private methods are non-virtual.
I will put `final` on methods parameters.

src/test/java/org/apache/commons/csv/CSVFormatSerializationTest.java
sman-81: Didn't know. Removed.

sman-81:
Thanks for your feedback, @garydgregory 
Changes implemented, please re-review.
TY!

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -2280,6 +2279,27 @@ private void validate() throws IllegalArgumentException {
...
+        if (duplicateHeaderMode == null) {
+            setFieldValue("duplicateHeaderMode",
+                    allowDuplicateHeaderNames ? DuplicateHeaderMode.ALLOW_ALL : DuplicateHeaderMode.DISALLOW);
+        }
+        return this;
+    }
+
+    private void setFieldValue(final String fieldName, final Object value) {
+        try {
+            Field fld = getClass().getDeclaredField(fieldName);
```
garydgregory, line 2294: Use final where you can.

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -2280,6 +2279,27 @@ private void validate() throws IllegalArgumentException {
...
+            fld.setAccessible(true); // make field non-final
+            fld.set(this, value);
+            fld.setAccessible(false);
+        } catch (Exception ex) {
```
garydgregory, line 2298: Unlikely to need to catch `Exception`, catching `ReflectiveOperationException` is cleaner IMO. But where is this used from anyway? If it's only called from `readResolve()` why not turn it into an `ObjectStreamException`?

src/test/java/org/apache/commons/csv/CSVFormatSerializationTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @ParameterizedTest(name = "[{index}] {0} ({1})")
+    @CsvSource(delimiterString = ";", value = {
+            "1.1.9;CSVFormat-Excel-1.9.0.ser"
+            //BROKEN ,"1.8;CSVFormat-Excel-1.8.ser" // field CSVFormat.delimiter was of type char (now String)
+            //BROKEN ,"1.7;CSVFormat-Excel-1.7.ser" // field CSVFormat.delimiter was of type char (now String)
+    })
+    void testDeserializeCsvFormat(String version, String fileName) throws IOException, ClassNotFoundException {
+
+        InputStream resourceAsStream = getClass().getResourceAsStream("/org/apache/commons/csv/serialization/" + fileName);
```
garydgregory, line 43: Use final where you can.

src/test/java/org/apache/commons/csv/CSVFormatSerializationTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.InputStream;
+import java.io.ObjectInputStream;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+/**
+ * Test deserialization of class {@link CSVFormat}.
+ */
+class CSVFormatSerializationTest {
```
garydgregory, line 33: Note that the convention in Commons CSV is that test classes and methods are public.

garydgregory (changes requested):
Hello @sman-81 
See comments. TY.

src/main/java/org/apache/commons/csv/CSVFormat.java
garydgregory: > In Java private methods are non-virtual. I will put `final` on methods parameters.

Of course, this comment applies only to parameters, just look at the rest of the code base ;-)

src/main/java/org/apache/commons/csv/CSVFormat.java
sman-81: The code in `try` might throw any of `NoSuchFieldException, SecurityException, IllegalArgumentException, IllegalAccessException`
`ObjectStreamException` is abstract, but a subclass can be used. `InvalidClassException` seems like the good choice.

src/test/java/org/apache/commons/csv/CSVFormatSerializationTest.java
sman-81: JUnit does not require that... I'll revert to public

garydgregory:
Hi @sman-81 
Shouldn't the `serialVersionUID` be updated since 1.10.0 adds `duplicateHeaderMode`? Wouldn't this be simpler?

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -356,11 +356,11 @@ public Builder setDelimiter(final String delimiter) {
         /**
          * Sets the duplicate header names behavior.
          *
-         * @param duplicateHeaderMode the duplicate header names behavior
+         * @param duplicateHeaderMode the duplicate header names behavior, may not be null
          * @return This instance.
          */
         public Builder setDuplicateHeaderMode(final DuplicateHeaderMode duplicateHeaderMode) {
-          this.duplicateHeaderMode = duplicateHeaderMode;
+          this.duplicateHeaderMode = Objects.requireNonNull(duplicateHeaderMode, "Mode required");
```
garydgregory, line 363: Why not use the `validate()` method like all other checks?

garydgregory:
Please see the discussion on the developer's mailing list https://lists.apache.org/thread/p017hp6qq1zcphg7pr1ko4swvvw8ynch
Mailing list info: https://commons.apache.org/mail-lists.html

sman-81:
Updated title of this PR to reflect the changed scope (all de/serialization logic is removed)

The PR offers the following fixes:

- CSVFormat.duplicateHeaderMode requires default DISALLOW for backward compatibility
- The field does not allow null values
- All de/serialization logic previously contained in this PR removed, commons-csv will effectively abandon serialization support
- Fix a bug in setNullString where member quotedNullString was inconsistently written (missing write in setQuote)
- Fix default of duplicateHeaderMode when set from boolean

PR #279, opened 2022-10-25
CSVParser: Identify duplicates in null, empty and blank header names

aherbert:
This changes the CSVParser to throw on a duplicate null header name when using DuplicateHeaderMode.DISALLOW. This matches the behaviour of CSVFormat.


garydgregory (approved):
LGTM

PR #295, opened 2022-12-27
Add support for trailing text after the closing quote, and EOF without a final closing quote, for Excel compatibility

DamjanJovanovic:
As per issue https://issues.apache.org/jira/browse/CSV-141 and based on what we did in Apache OpenOffice https://bz.apache.org/ooo/show_bug.cgi?id=126805

This adds a setting, allowTrailingText (for lack of a better name) that allows CSV fields to have trailing text after the closing quote, up to the next separator, which can contain anything except the separator character, and this extra text is appended as-is to the field contents (any further quoting is ignored). This is exactly how Excel behaves.

As this is a non-standard setting with surprising behaviour, I've made it off by default. Only CSVFormat.EXCEL has it on by default.

This doesn't fully fix CSV-141 yet as that has line ending issues too, but I'd like to investigate how Excel handles that first.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -206,6 +206,8 @@ public final class CSVFormat implements Serializable {
             return new Builder(csvFormat);
         }
 
+        private boolean allowEofWithoutClosingQuote;
+
         private boolean allowMissingColumnNames;
 
         private boolean allowTrailingText;
@@ -267,6 +269,7 @@ public final class CSVFormat implements Serializable {
             this.quotedNullString = csvFormat.quotedNullString;
             this.duplicateHeaderMode = csvFormat.duplicateHeaderMode;
             this.allowTrailingText = csvFormat.allowTrailingText;
+            this.allowEofWithoutClosingQuote = csvFormat.allowEofWithoutClosingQuote;
         }
 
         /**
@@ -288,6 +291,19 @@ public final class CSVFormat implements Serializable {
         @Deprecated
         public Builder setAllowDuplicateHeaderNames(final boolean allowDuplicateHeaderNames) {
             setDuplicateHeaderMode(allowDuplicateHeaderNames ? DuplicateHeaderMode.ALLOW_ALL : DuplicateHeaderMode.ALLOW_EMPTY);
+            return this;
+        }
+
+        /**
+         * Sets whether the last field on the last line, if quoted, can have no closing quote when the file ends, {@code true} if this is ok,
+         * {@code false} if {@link IOException} should be thrown.
+         *
+         * @param allowEofWithoutClosingQuote whether to allow the last field on the last line to have a missing closing quote when the file ends,
+         *                                    {@code true} if so, or {@code false} to cause an {@link IOException} to be thrown.
+         * @since 1.10.0
+         */
+        public Builder setAllowEofWithoutClosingQuote(final boolean allowEofWithoutClosingQuote) {
+            this.allowEofWithoutClosingQuote = allowEofWithoutClosingQuote;
             return this;
         }
 
@@ -827,7 +843,7 @@ public final class CSVFormat implements Serializable {
      * @see Predefined#Default
      */
     public static final CSVFormat DEFAULT = new CSVFormat(COMMA, DOUBLE_QUOTE_CHAR, null, null, null, false, true, CRLF, null, null, null, false, false, false,
-            false, false, false, DuplicateHeaderMode.ALLOW_ALL, false);
+            false, false, false, DuplicateHeaderMode.ALLOW_ALL, false, false);
 
     /**
      * Excel file format (using a comma as the value delimiter). Note that the actual value delimiter used by Excel is locale dependent, it might be necessary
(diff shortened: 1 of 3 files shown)


kinow (commented):
Code looks good to me! I think this could be made into a draft as it appears CSV-141 is still not fully fixed (the EOL, I see you pinged @garydgregory there too), and you said too that you want to investigate the behavior in Excel.

But looks like good progress, thanks!

src/test/java/org/apache/commons/csv/LexerTest.java
```diff
@@ -431,4 +431,14 @@ public void testTrimTrailingSpacesZeroLength() throws Exception {
         lexer.trimTrailingSpaces(buffer);
         assertThat(lexer.nextToken(new Token()), matches(EOF, ""));
     }
+
+    @Test
+    public void testTrailingTextAfterQuote() throws Exception {
```
garydgregory, line 436: Add a test asserting the expected behavior when if the new toggle is `false`.

garydgregory (changes requested):
@DamjanJovanovic 
Thank you for your PR.
A test is missing that an actual CSV file can be parsed and not parsed. See my other comment.

DamjanJovanovic:
Thank you, I've now added another patch, which adds a setting that controls whether the last field on the last line, if quoted, has to have a closing quote before the file ends. Also made both positive and negative tests for both patches.

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -206,8 +206,12 @@ public static Builder create(final CSVFormat csvFormat) {
             return new Builder(csvFormat);
         }
 
+        private boolean allowEOFWithoutClosingQuote;
```
garydgregory, line 209: Always use camel case allowEOFWithoutClosingQuote -> allowEofWithoutClosingQuote

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -288,6 +294,18 @@ public Builder setAllowDuplicateHeaderNames(final boolean allowDuplicateHeaderNa
             return this;
         }
 
+        /**
+         * Sets whether the last field on the last line, if quoted, can have no closing quote when the file ends, {@code true} if this is ok,
+         * {@code false} if {@link IOException} should be thrown.
+         *
+         * @param allowEOFWithoutClosingQuote whether to allow the last field on the last line to have a missing closing quote when the file ends,
+         *                                    {@code true} if so, or {@code false} to cause an {@link IOException} to be thrown.
+         */
```
garydgregory, line 303: Add Javadoc since tags to new public and protected elements.

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -1503,6 +1544,15 @@ public boolean getAllowDuplicateHeaderNames() {
         return duplicateHeaderMode == DuplicateHeaderMode.ALLOW_ALL;
     }
 
+    /**
+     * Gets whether the file can end before the last field on the last line, if quoted, has a closing quote.
+     *
+     * @return {@code true} if so, {@code false} to throw an {@link IOException}.
+     */
```
garydgregory, line 1551: Camel case.

garydgregory (changes requested):
Thank you for the update. See comments plus still missing an actual CSV file to test as previously mentioned.

src/main/java/org/apache/commons/csv/CSVFormat.java
DamjanJovanovic: Done, used 1.10.0.

src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -846,6 +881,8 @@ public CSVFormat getFormat() {
     public static final CSVFormat EXCEL = DEFAULT.builder()
```
garydgregory, line 881: I am wondering: Instead of this change, I wonder if we should have a new CSVFormat called `EXCEL_LENIENT`. Thoughts?

garydgregory: Hello? Anyone?

DamjanJovanovic: Maybe, but all spreadsheets are "lenient".

garydgregory:
I will find time to review over the weekend.

src/main/java/org/apache/commons/csv/CSVFormat.java
kinow: >Maybe, but all spreadsheets are "lenient".
 
Good point. I think @garydgregory suggested the EXCEL_LENIENT format to avoid changing the behavior for users of the EXCEL format. But if we are changing the format to what's supported in Excel, I guess that should be fine.

garydgregory:
I applied and then reverted this PR because it does not work. Notice the comment I made 2 weeks ago "A test is missing that an actual CSV file can be parsed and not parsed." This was never done, and sure enough, when I added a simple test based on the CSV file in the Jira description, it failed. Please see `CSVParserTest.testCSV141*` and in particular `testCSV141Excel`. You should:
- Rebase on git master
- Comment out of remove the `@Disabled` annotation for `testCSV141Excel` and see how the test fails when too much data is consumed.

If you choose to continue this work in another PR, make sure to add tests similar to `testCSV141Excel` to assert various permutations.

DamjanJovanovic:
> I applied and then reverted this PR because it does not work. Notice the comment I made 2 weeks ago "A test is missing that an actual CSV file can be parsed and not parsed." This was never done, and sure enough, when I added a simple test based on the CSV file in the Jira description, it failed.

It does work. The test you added is incorrect - Excel parses the file differently. Please see PR 303.

PR #303, opened 2023-01-23
Add support for trailing text after the closing quote, and EOF without a final closing quote, for Excel compatibility. Fix a unit test and add a RAT exclude for the sample CSV file.

DamjanJovanovic:
Continued from PR 295.

The test was failing because the test was wrong, not because my patches were wrong.

The test should match Excel's interpretation of the CSV file. Excel fuses lines 3 and 4 together, because the last field on line 3 doesn't end in a quote, so it continues into the next line. There, it stops at the initial quote, unquoting that portion, then also adds everything up to the comma, and all this becomes field 2 of line 3. The remaining fields on line 4 are interpreted as successive line 3 fields, and because the last field doesn't have a terminating quote, and the file ends in a new line, the last field also ends in a new line.

Once these corrections are made, the test passes.

Also add a RAT exclude for the sample file, which was missed out in commit 1269c133ff4637ed658fb8ab5d78a8671ecfed4a, and breaks the build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,6 +216,7 @@
             <excludes>
               <!-- These files are used as test data and test result specifications. -->
               <exclude>src/test/resources/org/apache/commons/csv/empty.txt</exclude>
+              <exclude>src/test/resources/org/apache/commons/csv/CSV-141/csv-141.csv</exclude>
               <exclude>src/test/resources/org/apache/commons/csv/csv-167/sample1.csv</exclude>
               <exclude>src/test/resources/org/apache/commons/csv/CSV-198/optd_por_public.csv</exclude>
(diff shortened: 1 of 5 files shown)


garydgregory:
"The test was failing because the test was wrong"

This is why I kept asking (twice) for file-based tests... please add file-based tests that cover use cases for these two new options beyond the one you changed. Let's make sure there are actual files that cover what Excel allows.

garydgregory:
@DamjanJovanovic 
- This functionality is now in git master, you are credited in `changes.xml`.
- The name of the new properties are different though: `setTrailingData(boolean)` and `setLenientEof(boolean)`.
- Closing this PR.

TY!

PR #309, opened 2023-02-23
(doc): Document duplicate header behavior

SethFalco:
A while back we discussed duplicate header behavior, and in the comments thought we should define what happens when one tries to get from a `CSVRecord`/`Map<String, String>` when this occurs.

Relevant comments:
* https://github.com/apache/commons-csv/pull/114#issuecomment-874321915
* https://github.com/apache/commons-csv/pull/114#issuecomment-876858602

I reviewed the code and manually tested to discover the current behavior:

* Adds two test cases that reinforce the current behavior.
* Updates the docstrings of `CSVRecord#get` and `CSVRecord#toMap` to state how this is handled.

---

It may be that instead of this, we may want to instead define a different behavior before we add documentation and tests to cover this case.

I'm honestly not sure what the expected behavior would be for this. I'm tempted to say that it should throw an exception in this instance. :thinking:

I'm not aware of the use-case for having duplicate headings in the first place, but I assume there is a reason for it. But I definitely don't see why one would use methods like `#get` or `#toMap` in cases that they know they have duplicate headers.

Maybe if the first and last index of the given header name differs, throw an exception?

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVRecord.java
+++ b/src/main/java/org/apache/commons/csv/CSVRecord.java
@@ -88,7 +88,7 @@ public final class CSVRecord implements Serializable, Iterable<String> {
     }
 
     /**
-     * Returns a value by name.
+     * Returns a value by name. If multiple instances of the header name exists, only the last occurence is returned.
      *
      * <p>
      * Note: This requires a field mapping obtained from the original parser.
@@ -311,7 +311,9 @@ public final class CSVRecord implements Serializable, Iterable<String> {
     }
 
     /**
-     * Copies this record into a new Map of header name to record value.
+     * Copies this record into a new Map of header name to record value. If multiple instances of a header name exists,
+     * only the last occurence is mapped.
+     *
      * <p>
      * Editing the map does not update this instance.
      * </p>
--- a/src/test/java/org/apache/commons/csv/CSVRecordTest.java
+++ b/src/test/java/org/apache/commons/csv/CSVRecordTest.java
@@ -23,6 +23,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertAll;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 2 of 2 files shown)


kinow (approved):
Hmmm, tricky. I think documenting the behavior sounds about right for now.

My go-to tool for parsing CSV ATM is Python. The stdlib `csv` module has the same behavior documented here, displaying the last occurrence.

```python
>>> import csv
>>> 
>>> with open('/tmp/test.csv') as f:
...   reader = csv.reader(f, delimiter=',')
...   for row in reader:
...     print(', '.join(row))
... 
A, A, B, B
1, 2, 5, 6
>>>
>>> with open('/tmp/test.csv') as f:
...   reader = csv.DictReader(f)
...   dict_from_csv = dict(list(reader)[0])
...   print(', '.join(list(dict_from_csv.keys())))
... 
A, B
>>>
>>> dict_from_csv
{'A': '2', 'B': '6'}
```

But whenever I need some serious CSV processing, then I use Pandas. With Pandas it automatically deduplicates the column names. Maybe that's a feature that we could have in Commons CSV too?

```python
>>> import pandas as pd
>>> 
>>> df = pd.read_csv('/tmp/test.csv')
>>> 
>>> ','.join(list(df.columns))
'A,A.1,B,B.1'
>>> 
>>> df['A']
0    1
Name: A, dtype: int64
>>> df['A.1']
0    2
Name: A.1, dtype: int64
>>> 
```

Cheers,
-Bruno

SethFalco:
I'm up for anything, but I'll at least give my personal thoughts on the methods described above. Note… I might be overcomplicated things tbh…

> The stdlib csv module has the same behavior documented here, displaying the last occurrence.

I feel just doing the last occurrence is very likely to lead to scenarios where:
1. User implements something using this library
2. They run into unexpected results due to duplicate header names
3. They code in a workaround to handle this situation as they feel best fits in with their application, likely either:
    * Disallow duplicate header names
    * Use a different method to iterate through columns
    * Deduplicate the headings themselves before providing it to Commons CSV.

I feel like throwing an exception makes more sense, so developers can skip to step 3 sooner. :thinking: 

> With Pandas it automatically deduplicates the column names

This actually crossed my mind, but I was hesitant to modify the data provided by users, at least it's probably best not to do this implicitly imo.

I'm also wary that this would screw up existing projects that might depend on/prefer simply allowing/disallowing duplicates. (i.e. want to allow duplicates and handle things through indexes / iteration, so this didn't cause a problem for them and therefore don't need the headers deduplicated)

If we want to put this in the library, I'd be tempted to do one of the following.

### `HeaderStrategy` Interface

Contains two functions:
* `#normalizeHeaders(headings)` - With given header names, output a new list that fits with whatever the strategy is going for.
* `#get(record, header)` - Fetch value(s) based on given column name.

The `#normalizeHeaders` function would be used during parsing. Implementations may choose not to do any normalization at all. (i.e. they don't want to rename headings)

This would allow things like deduplicating headings. I think it's best to separate normalization from fetching as implementations like this may want to display a warning in `#get` if the user tries to get one of the original header names before normalization. (i.e. `A,A` → `A.1,A.2` but user does `#get("A")`. People may be confused why it didn't work when it was indeed one of the original headings.)

The `#get` method would be used to get the value from the record. A strategy might be the:
* last occurrence
* the last occurrence with a non-empty value
* all values joined with a comma (`,`) or semicolon (`;`)

We could define one based on what we think the default behavior. This should reduce maintenance burden if users have different ideas on how this should be handled (so we don't have to support all use-cases ourselves). This enables consumers of the lib to override it with whatever they want.

### Repurpose `DuplicateHeaderMode`

We could introduce modes like `Deduplicate`. Unlike the above, this doesn't let users implement and provide their own strategies to the builder, they'd be stuck with our enum, which also means we shoulder the responsibility to implement whatever use-cases others have.)

### Something Else

Both solutions above might over-complicate things or increase maintenance effort for little gain. :thinking: Users already have the parser, from there they could technically do any normalization themselves already so long as we document one way or another what happens with duplicate headers if they do choose to leave them in.

garydgregory:
Hi All,

I'll start a thread on the ML for https://github.com/apache/commons-csv/pull/309#issuecomment-1441456258

PR #347, opened 2023-08-30
[CSV-147] Error message optimization during faulty CSV record read

gbidsilva:
This fix is related to : https://issues.apache.org/jira/browse/CSV-147.

If we have some faulty data in the CSV then the current error message which we are getting is something similar to below.
`java.io.IOException: (line 2) invalid char between encapsulated token and delimiter`

With this fix, what we will be getting is something similar to below,
`java.io.IOException: An error occurred while tying to parse the CSV content. Error in line: 2, position: 94, last parsed content: ...rec4,rec5,rec6,rec7,rec8`

**Update**
It has been decided to only to add the record position into the exception message and treat `getLastParsedContent` method as a new feature. Therefore this PR only contains the position related changes.


garydgregory:
I'm OK with adding the position but I am guessing someone will create a security issue for data exfiltration.

gbidsilva:
@garydgregory :
That is a good point. But IMO, whenever someone log the exception, at that point they should consider the security risks.
Isn't usually exception provide such details? as I remember I have came across such scenarios. 
What would be the best practice in this case ? - good to know :-)

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -438,6 +440,44 @@ public CSVParser(final Reader reader, final CSVFormat format, final long charact
...
+     * Example:
+     * </p>
+     * </p>
+     * If currently reading CSV record row contains following data and 'maxParsedTokenCount' is set to 5 and current reading position is col7
+     * </p>
+     * <pre>
+     * col1, col2, col3, col4, col5, col6, col7
+     * </pre>
+     * <p>
+     * then this would return following
```
elharo, line 458: would return --> returns

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -438,6 +440,44 @@ public CSVParser(final Reader reader, final CSVFormat format, final long charact
         this.recordNumber = recordNumber - 1;
     }
 
+    /**
+     * Return the parsed CSV content of current reading line up until this method is called.
```
elharo, line 444: "reading line" is awkward. Consider rephrasing

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -765,7 +805,15 @@ CSVRecord nextRecord() throws IOException {
         final long startCharPosition = lexer.getCharacterPosition() + this.characterOffset;
         do {
             this.reusableToken.reset();
-            this.lexer.nextToken(this.reusableToken);
+            // https://issues.apache.org/jira/browse/CSV-147
+            try {
+                this.lexer.nextToken(this.reusableToken);
+            } catch (IOException ioe) {
+                String errorMessage = "An exception occurred while tying to parse the CSV content. Issue in line: "
```
elharo, line 812: tying to parse --> parsing

elharo (changes requested):

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -1642,4 +1643,70 @@ private void validateRecordPosition(final String lineSeparator) throws IOExcepti
...
+                System.out.println(record.get(0) + " " + record.get(1) + " " + record.get(2) + " " + record.get(3)
+                        + " " + record.get(4) + " " + record.get(5) + " " + record.get(6) + " " + record.get(7)
+                        + " " + record.get(8) + " " + record.get(9));
+            }
+        });
+        String expectedErrorMessage = "Exception reading next record: java.io.IOException: An exception occurred " +
+                "while parsing the CSV content. Issue in line: 2, position: 94";
+        String actualMessage = exception.getMessage();
+        assertTrue(actualMessage.contains(expectedErrorMessage));
+        assertNotNull(csvParser);
```
elharo, line 1675: this assert doesn't seem helpful

elharo (changes requested):

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -765,7 +805,14 @@ CSVRecord nextRecord() throws IOException {
         final long startCharPosition = lexer.getCharacterPosition() + this.characterOffset;
         do {
             this.reusableToken.reset();
-            this.lexer.nextToken(this.reusableToken);
+            // https://issues.apache.org/jira/browse/CSV-147
```
garydgregory, line 808: No need for this comment IMO.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -765,7 +805,14 @@ CSVRecord nextRecord() throws IOException {
...
+            try {
+                this.lexer.nextToken(this.reusableToken);
+            } catch (IOException ioe) {
+                String errorMessage = "An exception occurred while parsing the CSV content. Issue in line: "
+                        + this.lexer.getCurrentLineNumber() + ", position: " + this.lexer.getCharacterPosition();
```
garydgregory, line 813: Simply message: "Exception parsing at line N, column C."

garydgregory (changes requested):
@gbidsilva 
Thank you for your updates. Please see my comments.

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -1642,4 +1643,70 @@ private void validateRecordPosition(final String lineSeparator) throws IOExcepti
...
+        CSVFormat csvFormat = CSVFormat.DEFAULT.builder()
+                .setHeader()
+                .setSkipHeaderRecord(true)
+                .build();
+
+        CSVParser csvParser = null;
+        try {
+            csvParser = csvFormat.parse(stringReader);
+        } catch (IOException e) {
+            fail("Failed to parse the CSV content");
```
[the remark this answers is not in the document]
gbidsilva: Done.

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -1642,4 +1643,70 @@ private void validateRecordPosition(final String lineSeparator) throws IOExcepti
...
+        }
+        final Iterable<CSVRecord> finalRecords = csvParser;
```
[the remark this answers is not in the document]
gbidsilva: Once above try catch is removed, we can simplify this.
Done.

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -1642,4 +1643,70 @@ private void validateRecordPosition(final String lineSeparator) throws IOExcepti
...
+        String csvContent = "col1,col2,col3,col4,col5,col6,col7,col8\n" +
+                "rec1,rec2,rec3,rec4,\"\"rec5\"\",rec6,rec7,rec8";
+
+        StringReader stringReader = new StringReader(csvContent);
+        CSVFormat csvFormat = CSVFormat.DEFAULT.builder()
+                .setHeader()
+                .setSkipHeaderRecord(true)
+                .build();
+
+        CSVParser csvParser = null;
```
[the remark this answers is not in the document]
gbidsilva: this is just to understand the usage and how the error message looks like for anyone looking at the test cases to figure the behavior.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -438,6 +440,44 @@ public CSVParser(final Reader reader, final CSVFormat format, final long charact
         this.recordNumber = recordNumber - 1;
     }
 
+    /**
+     * Return the parsed CSV content of currently parsed line up until this method is called.
```
elharo, line 444: If it's just a string I'm not sure it's "parsed". Maybe just "Return the text of the current line up until this method is called."

src/test/java/org/apache/commons/csv/CSVParserTest.java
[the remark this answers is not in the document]
elharo: I still see it. Do you need to push?

src/test/java/org/apache/commons/csv/CSVParserTest.java
elharo: csvParser is guaranteed non-null by the time this line is reached

src/main/java/org/apache/commons/csv/CSVParser.java
gbidsilva: Done.
However, column and position are two different things. Therefore kept the position as it is.
lexer is not having any method to access the column as I see.

src/main/java/org/apache/commons/csv/CSVParser.java
gbidsilva: done.

src/main/java/org/apache/commons/csv/CSVParser.java
gbidsilva: image

src/main/java/org/apache/commons/csv/CSVParser.java
gbidsilva: What we have in the buffer up until the error point is parsed as tokens. So I feel this is kind of OK.

src/test/java/org/apache/commons/csv/CSVParserTest.java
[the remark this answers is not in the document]
gbidsilva: removed.

gbidsilva:
@garydgregory @elharo :
Requested changes are done.

src/main/java/org/apache/commons/csv/CSVParser.java
garydgregory: > Done. However, column and position are two different things. Therefore kept the position as it is. lexer is not having any method to access the column as I see.

I meant position, sorry.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -438,6 +440,44 @@ public CSVParser(final Reader reader, final CSVFormat format, final long charact
...
+     * </pre>
+     * <p>
+     * then method returns following
+     * </p>
+     * <pre>
+     * col3, col4, col5, col6, col7
+     * </pre>
+     * @return parsed CSV content of current reading line
+     */
+    public String getLastParsedContent() {
```
garydgregory, line 465: I don't think this is needed anymore.

gbidsilva: @garydgregory : 
I think the whole point of this change is to provide more details on the error line in CSV to the user if there is any requirement.
In case if they want to know what has been parsed already and how it is parsed then they can use this method.
In fact, I think that is the original requirement of : https://issues.apache.org/jira/browse/CSV-147

garydgregory: I'll take another look later but, the fact that someone created a ticket does not make it a requirement for the component. As I've previously stated, adding position information is fine but not data IMO.

gbidsilva:
@garydgregory :
I understand your point. In this case, we will treat `getLastParsedContent` method as a new feature and decide whether it is actually needed or not in the product.
To make this PR work, I have removed the `getLastParsedContent` method content and only kept the positioning related changes.
Appreciate your feedback :-)

gbidsilva:
@garydgregory : let us know if there is anymore change to be done in this PR.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -765,7 +765,13 @@ CSVRecord nextRecord() throws IOException {
         final long startCharPosition = lexer.getCharacterPosition() + this.characterOffset;
         do {
             this.reusableToken.reset();
-            this.lexer.nextToken(this.reusableToken);
+            try {
+                this.lexer.nextToken(this.reusableToken);
+            } catch (IOException ioe) {
+                String errorMessage = "Exception during parsing at line: "
+                        + this.lexer.getCurrentLineNumber() + ", position: " + this.lexer.getCharacterPosition();
```
garydgregory, line 772: Wait a sec, why is code in the parser instead of in the lexer? All this catch does it extract information out of the lexer, so why not have exception thrown by the lexer provide this information in the first place in a better message?

gbidsilva: I will have a look on this...
Thanks for the review @garydgregory

gbidsilva: @garydgregory 
I had a look at what you are suggesting. Below are my thoughts...
If we look at the internals of method `Token nextToken(final Token token) throws IOException{}` in class `Lexer`, there are around 10 places where `IOException` can be thrown. 
Among them method calls to `parseEncapsulatedToken()`, `parseSimpleToken()` and `reader.read()` have high possibility of throwing parsing related `IOException`.

In that case we have following options.
1. handle each of method calls for customized IOException throwings with our enhanced exception message.
2. wrap `nextToken()` in class `Lexer` entirely and throw `IOException` with enhanced exception message.
3. keep the `this.lexer.nextToken(this.reusableToken)` wrapped with try catch in class `CSVParser` and rethrow IOException with enhanced message - this is what we have at the moment in PR.

IMO, keeping the try catch block in the current place we have at the moment seems more simple rather than handling exception in multiple places inside `Lexer` class.

Appreciate your opinion on this.

elharo: I vote for number 1. try to keep exception logic in one place.

src/main/java/org/apache/commons/csv/Lexer.java
```diff
@@ -367,8 +367,7 @@ private Token parseEncapsulatedToken(final Token token) throws IOException {
                         }
                         if (!Character.isWhitespace((char)c)) {
                             // error invalid char between token and next delimiter
-                            throw new IOException("(line " + getCurrentLineNumber() +
-                                    ") invalid char between encapsulated token and delimiter");
+                            throw new IOException("Invalid char between encapsulated token and delimiter at line: " + getCurrentLineNumber() + ", position: " + getCharacterPosition());
```
elharo, line 370: This probably shouldn't be an IOException but that issue is not new with this PR

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -1642,4 +1642,26 @@ private void validateRecordPosition(final String lineSeparator) throws IOExcepti
...
+                "rec1,rec2,rec3,rec4,rec5,rec6,rec7,rec8,\"\"rec9\"\",rec10";
+
+        StringReader stringReader = new StringReader(csvContent);
+        CSVFormat csvFormat = CSVFormat.DEFAULT.builder()
+                .setHeader()
+                .setSkipHeaderRecord(true)
+                .build();
+
+        CSVParser csvParser = csvFormat.parse(stringReader);
+        Exception exception = assertThrows(UncheckedIOException.class, () -> {
```
elharo, line 1657: UncheckedIOException is not right either, but again not new in this PR

elharo (approved):

gbidsilva:
@garydgregory : Checkstyle issue has been fixed.

gbidsilva:
@garydgregory : Anything pending from development side for this to be merged ?

gbidsilva:
> @gbidsilva Thank you for your updates. Please see my comments.

Completed.

elharo (approved):

garydgregory (approved):
LGTM

PR #412, opened 2024-04-05
MongoDB CSV empty first column parsing fix

ikfid:
when printed a record with an empty first column with MongoDBCsv predefined format, the csv file cannot be then parsed with the same MongoDBCsv predefined format.

this is due to the conflict between 'escape' and 'quote' character branches in Lexer. They are both set to ' " ' for MongoDBCsv.

fix: in Lexer reordered 'escape' and 'quote' branches.

### DIFF
--- a/src/main/java/org/apache/commons/csv/Lexer.java
+++ b/src/main/java/org/apache/commons/csv/Lexer.java
@@ -334,18 +334,7 @@ final class Lexer implements Closeable {
         while (true) {
             c = reader.read();
 
-            if (isEscape(c)) {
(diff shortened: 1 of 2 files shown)


garydgregory:
TY @ikfid for your PR. Merged. Please note the changes I made to the test to convert it into a proper JUnit test; there is no need to handle exceptions "manually": https://github.com/apache/commons-csv/commit/70c9418e71c32a6a8248f9b2e62aad6b3892a485

ikfid:
@garydgregory , thanks !) the proper test is way better :)

PR #439, opened 2024-07-04
Fix documentation

sigee:


CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@
 Contributing to Apache Commons CSV
 ======================
 
-You have found a bug or you have an idea for a cool new feature? Contributing code is a great way to give something back to
+You have found a bug, or you have an idea for a cool new feature? Contributing code is a great way to give something back to
```
aherbert, line 44: This is auto generated text so this is not the place to fix this. Can you revert please.

pom.xml
```diff
@@ -404,6 +404,13 @@
       </plugin>
     </plugins>
   </reporting>
+  <repositories>
```
aherbert, line 407: This repository is already included from the parent pom: org.apache:apache

pom.xml
```diff
@@ -44,7 +44,7 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>4.11.0</version>
+      <version>5.11.0</version>
```
aherbert, line 47: I am not sure why you are upgrading this dependency; the pr has no use of this plugin.

aherbert (changes requested):

src/main/java/org/apache/commons/csv/Lexer.java
```diff
@@ -22,6 +22,8 @@
 import java.io.Closeable;
 import java.io.IOException;
 
+import org.apache.commons.io.IOUtils;
```
aherbert, line 25: We do not require this import for the code. You can use it directly in the javadoc.

src/main/java/org/apache/commons/csv/Lexer.java
```diff
@@ -467,7 +469,7 @@ boolean readEndOfLine(int ch) throws IOException {
      * On return, the next character is available by calling {@link ExtendedBufferedReader#getLastChar()}
      * on the input stream.
      *
-     * @return the unescaped character (as an int) or {@link Constants#EOF} if char following the escape is
+     * @return the unescaped character (as an int) or {@link IOUtils#EOF} if char following the escape is
```
aherbert, line 472: ```{@link org.apache.commons.io.IOUtils#EOF IOUtils.EOF}```

aherbert (changes requested):

pom.xml
garydgregory: Worse, it's not compatible with Java 8 🙁

garydgregory:
Closing, see comments.

PR #442, opened 2024-07-09
Fix some Javadoc links

sigee:


garydgregory:
Hello @sigee 
Thank you for your PR. 
I fixed this using imports instead of FQCNs.
You are credited in `changes.xml`.
Closing

PR #443, opened 2024-07-09
Remove unnecessary boxing/unboxing

sigee:


garydgregory:
-1: Nope, sorry. I feel it is best to be explicit to see where boxing and unboxing happens in a lower level library. This matters especially when relying on unboxing which lead to NPEs. This can also help in writing more performance oriented code when you see boxing and unboxing happen over and over.

sigee:
Sorry, but I disagree.

> This matters especially when relying on unboxing which lead to NPEs.

The foolowing code snipper use auto unboxing and lead to NPE.

```
public class MyTest {
    public static void main(String[] args) {
        Integer a = null;
        Integer b = null;
        Integer c = add(a, b);

        System.out.println(c);
    }


    private static int add(int a, int b) {
        return a + b;
    }
}
```

If I use explicit unboxing, the result is the same. There is the same NPE at the same place.

```
public class MyTest {
    public static void main(String[] args) {
        Integer a = null;
        Integer b = null;
        Integer c = add(a.intValue(), b.intValue());

        System.out.println(c);
    }


    private static int add(int a, int b) {
        return a + b;
    }
}

```

> This can also help in writing more performance oriented code when you see boxing and unboxing happen over and over.

You can't avoid bad performing code in the external system no matter what. You can make good performing code inside the library by make sure your API is well designed and you don't use the "same" parameter differently in different methods.
If performance matters for you, you should use primitive if possible.
For example what is the reason inside the `Constants`  to use `Character` as `DOUBLE_QUOTE_CHAR` instead of using primitive `char`? It is a constant value which is not `null`, so every time it is a double quote. You should use wrapper (`Character`) only if you want to handle null values. So in this case you could save heap operations (do better performance) by using primitiv `char` constant.

If you write explicit (un)boxing, that is a transformation. If you miss it for some reasons, java will do it only if needed. But there is no way to spare any transformation by using explicit (un)boxing in specific places. The only way to spare them by designing consistent API that uses only primitives or only wrappers, but not mixing them (if possible), but writeing explicitly (un)boxing won't help make better performance, just less readable code. IMHO.

garydgregory:
FWIW the title of the PR is wrong. You can't remove [un]boxing, the complied generates it whether you like it or not.

When you see a de-reference like a.b() you know by just glancing at the code that you're in trouble if a is null. If you hide the unboxing, you don't.

sigee:
> You can't remove [un]boxing, the complier generates it whether you like it or not.

Yes the PR is about to remove the **explicit** (un)boxing, because it is not necessary. The compiler generates it.

> When you see a de-reference like a.b() you know by just glancing at the code that you're in trouble if a is null. If you hide the unboxing, you don't.

I get your point in case of unboxing/de-referencing, but what is the reason behind the **explicit** boxing?

garydgregory:
Explicit boxing shows where you are creating an Object or getting one from a cache (from the Integer cache for example). It is also symmetric in style.

garydgregory:
Closing: Won't do.

PR #456, opened 2024-08-06
[CSV-293] Add support for multiple null String values

sigee:
This change satisfies the https://issues.apache.org/jira/browse/CSV-293 issue to support multiple null String values.
- I changed the CSVFormat class to be able to store not only one String representing the null value, but an array of Strings for multiple values.
- I also changed the Builder to handle the array of Strings as null values.
- To be backward compatible, I left the old setNullString(String) API untouched but changed the behavior to store the String in an array as a single value.
- I changed the equals body to use Arrays.equals instead of Object.equals since the String nullString changed to a String[] nullStrings.
- I also changed the calculation of quotedNullString in the first commit, but I completely removed it in the second commit, because It was calculated and assigned in a bunch of places, but was used only in one place, so left the minimum required logic in that print() method.
- Finally, I changed the CSVParser to use the null values set in the CSVFormat.
- Testing:
  - Since the changes are backward compatible no test changed, only the toString as NullString became NullStrings.
  - I created test cases in the issues/JiraCsv293Test for the new multiple null values in case of
    - multiple values are provided
    - one value is provided
    - null is provided
  - I also wrote test cases for the original API.
  - I also write a test for the new getter.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -225,11 +225,11 @@ public final class CSVFormat implements Serializable {
 
         private boolean ignoreSurroundingSpaces;
 
-        private String nullString;
+        private String[] nullStrings;
 
         private Character quoteCharacter;
 
-        private String quotedNullString;
+        private String[] quotedNullStrings;
 
         private QuoteMode quoteMode;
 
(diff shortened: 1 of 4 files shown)


garydgregory:
This PR has an empty description.

sigee:
@garydgregory I wrote a detailed description.

garydgregory:
Hello @sigee 

Please see my [comment](https://issues.apache.org/jira/browse/CSV-293?focusedCommentId=17872149&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17872149) in the Jira ticket.

PR #473, opened 2024-09-10
Replaced BufferedReader without synchronization blocks

mmstan:
Replaced the BufferedReader from java.io package by a modified BufferedReader from the Apache Harmany project.

All synchronized blocks are removed from the BufferedReader in order to reach a better performance since in Java 15 the synchronisation optimization is deprecated and in Java 18 the optimization is removed.

I added a jmh file for performance test results.

[jmh-result.json](https://github.com/user-attachments/files/16940825/jmh-result.json)

### DIFF
--- a/src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
+++ b/src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
@@ -22,7 +22,6 @@ import static org.apache.commons.csv.Constants.LF;
 import static org.apache.commons.csv.Constants.UNDEFINED;
 import static org.apache.commons.io.IOUtils.EOF;
 
-import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.Reader;
 
@@ -35,7 +34,7 @@ import org.apache.commons.io.IOUtils;
  * {@link #read()}. This reader also tracks how many characters have been read with {@link #getPosition()}.
  * </p>
  */
-final class ExtendedBufferedReader extends BufferedReader {
+final class ExtendedBufferedReader extends BufferedReaderFork {
 
     /** The last char returned */
     private int lastChar = UNDEFINED;
--- a/pom.xml
+++ b/pom.xml
@@ -189,11 +189,11 @@
     <!-- JaCoCo: Don't make code coverage worse than: -->
     <commons.jacoco.haltOnFailure>true</commons.jacoco.haltOnFailure>
     <commons.jacoco.classRatio>1.00</commons.jacoco.classRatio>
-    <commons.jacoco.instructionRatio>0.98</commons.jacoco.instructionRatio>
-    <commons.jacoco.methodRatio>1.00</commons.jacoco.methodRatio>
-    <commons.jacoco.branchRatio>0.97</commons.jacoco.branchRatio>
-    <commons.jacoco.lineRatio>0.99</commons.jacoco.lineRatio>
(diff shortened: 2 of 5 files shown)


src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -35,7 +34,7 @@
  * {@link #read()}. This reader also tracks how many characters have been read with {@link #getPosition()}.
  * </p>
  */
-final class ExtendedBufferedReader extends BufferedReader {
+final class ExtendedBufferedReader extends BufferedReaderFork {
```
ppkarwasz, line 37: @mmstan,

If your goal is to use a `Reader` that does not use synchronized blocks, why don't you replace `BufferedReader` with `Reader`? As far as I can tell, `ExtendedBufferedReader` implements almost all the methods used by `Lexer`.

`BufferedReaderFork` is rather a no-go: it would require a lot of maintenance work over the years, whereas classes such as `Reader` and `BufferedReader` are maintained by OpenJDK.

garydgregory:
This PR should be put in the "draft" state for now IMO.

mmstan:
@garydgregory @ppkarwasz 

I checked the proposal without BufferedReader and I extended the ExtendedBufferedReader from Reader.
Unfortunately this solution is still slower as before in my use case.
And the UT failed because mark() and reset() are not implemented in ExtendedBufferedReader and Reader.

ppkarwasz:
> @ppkarwasz The ExtendedBufferedReader calls reset() and mark() methods, we cannot remove the calls. I think it is not possible to extend the ExtendedBufferedReader from Reader, the implementations for reset() and mark() are missing. Or I missunderstand your proposal.

No, you didn't misunderstand. Effectively, looking at the code, the task is larger than I expected, so you'll need to add some methods to `ExtendedBufferedReader` if you refactor it as direct derivative class of `Reader`. Please, provide a **minimal** implementation: if `Lexer` does not need a method, don't implement it, because it will be dead code.

In your current approach you copied and refactored the entire `BufferedReader` class. `Lexer` does not need all of that.

garydgregory:
Let's do this through Commons IO where I added `UnsynchronizedBufferedReader` and `UnsynchronizedReader`, such that `ExtendedBufferedReader` can extend `UnsynchronizedBufferedReader`. 

On Java 8 on Windows 10, running `PerformanceTest` a few times and picking best runs, shows an improvement from:
```
Best time out of 10 is 1,290 milliseconds.
```
to:
```
Best time out of 10 is 1,023 milliseconds.
```
or ~20% improvement

I push an RC for Commons IO in the coming week or so, then we can pick it up here.

garydgregory:
@mama-btc-es 
CC: @ppkarwasz 

Please test git master which now uses an unsynchronized reader.

garydgregory:
Ping @mama-btc-es

mama-btc-es:
@garydgregory 
I checked the changes on git master. Now my use case is so fast as before. Thank you.

garydgregory:
@mama-btc-es 
TY for your feedback. 
OK to close this ticket then or do see any more changes needed?

mama-btc-es:
@garydgregory 
No changes from my point of view anymore. You can close the ticket.
I will delete my fork branch and the PR.

garydgregory:
Closing after review and agreement from @mama-btc-es 
TY!

PR #487, opened 2024-10-11
Update maven.yml remove java versions

pabloacinass:


aherbert:
It is sensible to test a well used library against all LTS version of Java. You may not use Java 8, but others may do.

garydgregory:
-1, I agree with @aherbert. We build to the Java 8 target.

PR #502, opened 2024-11-07
CSV-196-TrackBytePositions

DarrenJAN:
Add support in Commons CSV for tracking byte positions during parsing.

Summary of Modifications

1. **Test Data Files**: Added new test data files, and updated pom.xml to exclude these files from RAT checks, avoiding unapproved license checks.
2. **CSVParser class**:
 Constructor Enhancements
a. Added support for an optional parameter -- **String encoding**--, which specifies the encoding to use for the reader.
4. **CSVRecord class**
- private long characterByte:  start byte position of this record 
Add new Constructor: support track byte positions in record class 
5. **ExtendedBufferedReader Class**:

- private long bytesRead: Tracks the number of bytes read so far.
- private long bytesReadMark: Stores the marked byte position.
- CharsetEncoder encoder: Encoder used to calculate byte size of characters.
- getCharBytes(int current): This function calculates character bytes based on UTF-8 encoding. Note: it only supports UTF-8 due to the encoding algorithm used. Full encoding can be supported and we just need more effort on this. 
- reset() and mark() Methods: Enhanced to prevent consuming characters and bytes unintentionally.

Test result:
mvn 
image
image
Pass unit tests and other restrictions


src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -2097,6 +2097,30 @@ public CSVParser parse(final Reader reader) throws IOException {
...
+     *
+     * @param reader the input stream
+     * @param characterOffset the character offset to start parsing from
+     * @param recordNumber the initial record number to start counting from
+     * @param encoding the character encoding of the input stream
+     * @return a parser over a stream of {@link CSVRecord}s.
+     * @throws IOException If an I/O error occurs
+     * @throws CSVException Thrown on invalid input.
+     */
+    public CSVParser parse(final Reader reader, final long characterOffset, final long recordNumber, String encoding) throws IOException {
```
garydgregory, line 2120: No new public constructors please. You can augment the builder and add a private constructor.

All new and protected elements should have a Java `@since 1.13.0`.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -511,10 +511,39 @@ public CSVParser(final Reader reader, final CSVFormat format) throws IOException
...
+     *            The next record number to assign
+     * @param encoding
+     *            The encoding to use for the reader
+     * @throws IllegalArgumentException
+     *             If the parameters of the format are inconsistent or if either the reader or format is null.
+     * @throws IOException
+     *             If there is a problem reading the header or skipping the first record
+     * @throws CSVException Thrown on invalid input.
+     */
+    public CSVParser(final Reader reader, final CSVFormat format, final long characterOffset, final long recordNumber,
```
garydgregory, line 541: See previous comment.

src/main/java/org/apache/commons/csv/CSVRecord.java
```diff
@@ -144,6 +159,15 @@ public long getCharacterPosition() {
         return characterPosition;
     }
 
+    /**
+     * Returns the start byte of this record as a character byte in the source stream.
```
garydgregory, line 163: Please follow the same Javadoc patterns as in other getter methods: A getter "Gets...".

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -49,13 +53,27 @@ final class ExtendedBufferedReader extends UnsynchronizedBufferedReader {
     private long position;
     private long positionMark;
 
+    /** The number of bytes read so far */
+    private long bytesRead;
+    private long bytesReadMark;
+
+    /** Encoder used to calculate the bytes of characters */
+    CharsetEncoder encoder;
```
garydgregory, line 61: Make this new instance variable private.

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -49,13 +53,27 @@ final class ExtendedBufferedReader extends UnsynchronizedBufferedReader {
...
+
     /**
      * Constructs a new instance using the default buffer size.
      */
     ExtendedBufferedReader(final Reader reader) {
         super(reader);
     }
 
+    ExtendedBufferedReader(final Reader reader, String encoding) {
```
garydgregory, line 70: Use a `Charset` instead of a Charset name.

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -118,11 +137,43 @@ public int read() throws IOException {
...
         lastChar = current;
         position++;
         return lastChar;
     }
 
+    /**
+     *  In Java, a char data type are based on the original Unicode
+     *  specification, which defined characters as fixed-width 16-bit entities.
+     *   U+0000 to U+FFFF:
+     *     - BMP, represented using 1 16-bit char
```
garydgregory, line 152: Use Javadoc lists `<ol>` or `<ul>` if you expect the formatting to be maintained.

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -118,11 +137,43 @@ public int read() throws IOException {
...
+     *     - Consists of UTF-8 1-byte, 2-byte, some 3-byte chars
+     *   U+10000 to U+10FFFF:
+     *     - Supplementary characters, represented as a pair of characters,
+     *     the first char from the high-surrogates range (\uD800-\uDBFF),
+     *     and the second char from the low-surrogates range (uDC00-\uDFFF).
+     *     - Consists of UTF-8 some 3-byte chars and 4-byte chars
+     */
+    private long getCharBytes(int current) throws CharacterCodingException {
+        char cChar = (char) current;
```
garydgregory, line 161: Use `final` where possible.

src/test/java/org/apache/commons/csv/JiraCsv196Test.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.Reader;
+
+
+import org.junit.jupiter.api.Test;
+
+
+public class JiraCsv196Test {
+    @Test
+    public void parseThreeBytes() throws IOException {
+
```
garydgregory, line 32: No need for extra blank line.

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -701,6 +701,84 @@ public void testGetHeaderComment_NoComment3() throws IOException {
...
+        assertEquals(3, record.getRecordNumber());
+        assertEquals(code.indexOf('2'), record.getCharacterPosition());
+        assertEquals(record.getCharacterByte(), 95);
+
+        assertNotNull(record = parser.nextRecord());
+        assertEquals(4, record.getRecordNumber());
+        assertEquals(code.indexOf('3'), record.getCharacterPosition());
+        assertEquals(record.getCharacterByte(), 154);
+
+        parser.close();
```
garydgregory, line 741: Use a try-with-resources block instead of manually closing a resource.

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -701,6 +701,84 @@ public void testGetHeaderComment_NoComment3() throws IOException {
...
     }
 
+    @Test
+    public void testGetRecordThreeBytesRead() throws Exception {
+        String code = "id,date,val5,val4\n" +
+            "11111111111111,'4017-09-01',きちんと節分近くには咲いてる～,v4\n" +
+            "22222222222222,'4017-01-01',おはよう私の友人～,v4\n" +
+            "33333333333333,'4017-01-01',きる自然の力ってすごいな～,v4\n";
+        // String code = "'1',4";
+        // final CSVFormat format = CSVFormat.newFormat(',').withQuote('\'');
```
garydgregory, line 711: Are these comments needed?

src/main/java/org/apache/commons/csv/CSVRecord.java
```diff
@@ -144,6 +159,15 @@ public long getCharacterPosition() {
...
+     *
+     * @return the start byte of this record as a character byte in the source stream.
```
garydgregory, line 165: Either the code or the comment is wrong. The return value is a `long` but the comment talks about a `byte`. It would help to make the comment clear as to why the return type is a `long` if that is indeed correct.

src/main/java/org/apache/commons/csv/CSVRecord.java
```diff
@@ -48,6 +48,11 @@ public final class CSVRecord implements Serializable, Iterable<String> {
      */
     private final long characterPosition;
 
+    /**
+     * The start byte of this record as a character byte in the source stream.
```
garydgregory, line 52: This comment is confusing because it talks about a `byte` but the type is a `long`. It would be better to explain the mismatch, if that is indeed the intent.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -511,10 +511,39 @@ public CSVParser(final Reader reader, final CSVFormat format) throws IOException
...
+        String encoding) throws IOException {
```
garydgregory, line 542: Use a `Charset`, not a charset name.

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -49,13 +53,27 @@ final class ExtendedBufferedReader extends UnsynchronizedBufferedReader {
     private long position;
     private long positionMark;
 
+    /** The number of bytes read so far */
```
garydgregory, line 56: A sentence ends in a period `.`.

garydgregory (changes requested):
Hello @DarrenJAN 

Thank you for providing a PR!

Please see my comments. Note that some comments will apply to more to one location in the code.

src/main/java/org/apache/commons/csv/CSVFormat.java
DarrenJAN: I noticed that there is only builder class in CSVFormat, there are no builder class in CSVParser, would you like me to add one?
There are CSVParserBuilder() from com.opencsv.CSVParser, that is another package.

garydgregory: Uh? See https://github.com/apache/commons-csv/blob/96427fc72fe5873fd0bd23cac147eef302bf5b75/src/main/java/org/apache/commons/csv/CSVParser.java#L151

DarrenJAN: It seems like I did not rebase on the master branch. Thanks.

garydgregory: This section is not resolved.

DarrenJAN:
Hi Gary,
I submitted another commit. Here are the changes that I made:

1.  Add augment the builder and add a private constructor to create CVSParser class
2. Delete unneeded constructors in CSVformat class 
3. Use a try-with-resources block
4. Improve the comments and fix the indentation
5. Use a Charset instead of a string Charset name in ExtendedBufferedReader class

Thanks
Yuzhan

src/main/java/org/apache/commons/csv/CSVRecord.java
DarrenJAN: We are calculating the number of bytes required to encode a single character and storing this value as a `long`. This choice provides a larger range for the position and ensures consistency with the design, as position is also defined as a `long`.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -153,6 +153,7 @@ public static class Builder extends AbstractStreamBuilder<CSVParser, Builder> {
         private CSVFormat format;
         private long characterOffset;
         private long recordNumber = 1;
+        private Charset charset;
```
garydgregory, line 156: You don't need this, it's already in the superclass.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -164,7 +165,7 @@ protected Builder() {
         @SuppressWarnings("resource")
         @Override
         public CSVParser get() throws IOException {
-            return new CSVParser(getReader(), format != null ? format : CSVFormat.DEFAULT, characterOffset, recordNumber);
+            return new CSVParser(getReader(), format != null ? format : CSVFormat.DEFAULT, characterOffset, recordNumber, charset);
```
garydgregory, line 168: Call getCharset()

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -200,6 +201,16 @@ public Builder setRecordNumber(final long recordNumber) {
             return asThis();
         }
 
+        /**
+         * Sets the character encoding to be used for the reader.
+         *
+         * @param charset the character encoding.
+         * @return this instance.
+         */
+        public Builder setCharset(final Charset charset) {
```
garydgregory, line 210: You don't need this, see above.

garydgregory (changes requested):
Hello @DarrenJAN 

Thank you for your updates. There is one adjustment to make where code is not needed because it is provided in the superclass.

DarrenJAN:
Hi Gary, 

Just submit another change.

Please note that due to the encoding algorithm used, it only supports UTF-8 for now. Full encoding can be supported; I just need to put more effort into this.

Thanks
Yuzhan

garydgregory (changes requested):
Hello @DarrenJAN 

Now that I look at this again, we are always using the Charset, which is not enough to only have the feature enabled as an opt-in. This is due to the Charset being used to read the file in the first place, it's likely that the Charset will be often set, so I think we need a boolean flag that says "recordByteCount". The flag should drive whether the feature is enabled or not, not the Charset. You still need the Charset, but that's not what enables the feature.

Does that make sense? I think adding a boolean recordByteCount to the builder is what's needed here. WDYT?

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
garydgregory: Not resolved.

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
garydgregory: Not resolved.

garydgregory (changes requested):
Hello @DarrenJAN 

There are many comments that were incorrectly marked resolved. See also my comment about adding a boolean to drive the new opt-in behavior.

DarrenJAN:
Hi Gary, 

In ExtendedBufferedReader.java,  this function getCharBytes(int current),  the logic of this code is based on the UTF-8 encoding specification.  We need extra implementation to support the full set of characters. 

"Hence, the need for a boolean or some other type that's not a Charset, maybe even an Enum if there a need for something more than on and off."
--- We used a String here. We only did a simple check whether String is null to enable this encoder feature since our customers only asked for UTF-8 encoder. I agree with adding a flag to control this feature.

DarrenJAN:
Hi Gary,  any thoughts on this?

garydgregory:
Hello @DarrenJAN 
Yes, follow up on my comments to add a boolean feature toggle 😉

DarrenJAN:
Hi @garydgregory,
I submitted another commit to add a boolean flag

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -153,6 +153,7 @@ public static class Builder extends AbstractStreamBuilder<CSVParser, Builder> {
         private CSVFormat format;
         private long characterOffset;
         private long recordNumber = 1;
+        private boolean enableByteTracking = false;
```
garydgregory, line 156: Don't initialize to the default value.

src/main/java/org/apache/commons/csv/CSVParser.java
```diff
@@ -200,6 +201,11 @@ public Builder setRecordNumber(final long recordNumber) {
             return asThis();
         }
 
+        public Builder setEnableByteTracking(final boolean enableByteTracking) {
```
garydgregory, line 204: Javadoc missing.

src/main/java/org/apache/commons/csv/CSVRecord.java
garydgregory: ping

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -49,13 +53,27 @@ final class ExtendedBufferedReader extends UnsynchronizedBufferedReader {
...
+    private CharsetEncoder encoder;
+
     /**
      * Constructs a new instance using the default buffer size.
      */
     ExtendedBufferedReader(final Reader reader) {
         super(reader);
     }
 
+    ExtendedBufferedReader(final Reader reader, Charset charset, boolean enableByteTracking) {
```
garydgregory, line 70: Javadoc missing

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -118,11 +137,53 @@ public int read() throws IOException {
             current == EOF && lastChar != CR && lastChar != LF && lastChar != EOF) {
             lineNumber++;
         }
+        if (encoder != null) {
```
garydgregory, line 140: Test is missing `enableByteTracking`.

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -118,11 +137,53 @@ public int read() throws IOException {
...
+            this.bytesRead += getCharBytes(current);
+        }
         lastChar = current;
         position++;
         return lastChar;
     }
 
+    /**
+     * In Java, the {@code char} data type is based on the original Unicode
```
garydgregory, line 149: Getter get, IOW, the Javadoc should start with "Gets ...".

src/main/java/org/apache/commons/csv/Lexer.java
```diff
@@ -103,6 +103,15 @@ long getCharacterPosition() {
         return reader.getPosition();
     }
 
+    /**
+     * Returns the number of bytes read
```
garydgregory, line 107: "Returns" -> "Gets".

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -701,6 +701,77 @@ public void testGetHeaderComment_NoComment3() throws IOException {
...
+            assertEquals(4, record.getRecordNumber());
+            assertEquals(code.indexOf('3'), record.getCharacterPosition());
+            assertEquals(record.getCharacterByte(), 154);
+        };
+
+    }
+
+    @Test
+    public void testGetRecordFourBytesRead() throws Exception {
+        String code = "id,a,b,c\n" +
```
garydgregory, line 743: Use `final`.

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -701,6 +701,77 @@ public void testGetHeaderComment_NoComment3() throws IOException {
         }
     }
 
+    @Test
+    public void testGetRecordThreeBytesRead() throws Exception {
```
garydgregory, line 705: Use `final`.

garydgregory (changes requested):
Hello @DarrenJAN 

Thank you for your updates. Please see my comments.

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
DarrenJAN: fixed

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
DarrenJAN: `enableByteTracking` will take control on `encoder`, so no extra testing is needed

src/main/java/org/apache/commons/csv/CSVRecord.java
```diff
@@ -48,6 +48,11 @@ public final class CSVRecord implements Serializable, Iterable<String> {
...
+     */
+    private final long characterByte;
```
garydgregory, line 54: This name is too confusing IMO, please rename to `bytePosition` to mirror the existing `characterPosition`.

garydgregory (changes requested):
Hello @DarrenJAN 

Thank you for updates.

I think we have the basic logic done but I have some additional change requests.

TY!

DarrenJAN:
Hi @garydgregory, 

Happy Holidays!  The changes that I made:
1. New private elements do not need a Javadoc since tag.
2. Return type of **getCharByte** is int
3. Change **getCharBytes** to **getEncodedCharLength**
4. Add Javadoc @param for enableByteTracking 
5. CSVRecord:  Remove the old constructors and adapt the existing test.
6. Change **characterByte** to **bytePosition** @yunzvanessa How do you think this change? 


Test:
mvn 
image

garydgregory (changes requested):
Hello @DarrenJAN

Thank you for updating the PR.

You missed a couple of my comments regarding documentation.

Please rebase on git master to pickup updates that will automatically show you where you missed using `final` and other Checkstyle updates. You'll have to resolve conflicts in one test file.

TY & happy hols!

garydgregory (changes requested):
Hello @DarrenJAN 
I cannot edit files in this PR, which is unusual; otherwise, I would have added the missing Javadoc `@since` tags. Please do so and see my unresolved comments.
TY!

DarrenJAN:
Hi @garydgregory, I checked and made the enhancements:
1. git rebase master
2. For new public and protected methods, add javadoc `since 1.13.0`
3. Clarify the comment:

Test:
`mvn `
image

garydgregory:
https://issues.apache.org/jira/browse/CSV-196

DarrenJAN:
Hi @garydgregory , 

Thank you very much. When is the next release?

Yuzhan

garydgregory:
Hello @DarrenJAN 
I have three open release candidates ATM, so when that's done, I'll start the release process for this component. In the meantime, please test a build from our Maven snapshot repository.

PR #519, opened 2025-01-19
Use mockito version from commons-parent

raboof:
replaces https://github.com/apache/commons-csv/pull/514

this seems small enough not to need an entry in `changes.xml`, right?

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [x] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,7 +37,7 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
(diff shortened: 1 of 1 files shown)


garydgregory:
@raboof 
I agree! Merged.

PR #520, opened 2025-01-19
Attempt to fix javadoc generation warning/fail

slachiewicz:
It looks like the links section is problematic Adding a reference to Javadocs 21 generates the warning/fail.


garydgregory:
I'm not sure that's the solution anyway. The problem is that Javadoc doesn't know about the module name, so we probably need to point it to the module we generate at build time.

garydgregory:
But you have links to everything like the JDK classes?

garydgregory:
Closing: This is now fixed in the parent POM. TY both 👍

PR #605, opened 2026-06-05
Fix surrogate pair byte counting in ExtendedBufferedReader array read

digi-scrypt:
1. with byte tracking on (setTrackBytes + a charset), read(char[]) advances lastChar to the last buffer char before counting, and the per-char helper reads that field instead of the actual preceding char, so a surrogate pair gets matched against the wrong neighbor (the loop also ran to length instead of offset+length).
2. a 4-byte char taken through the char[] path, e.g. a multi-character delimiter holding a supplementary character, then throws CharacterCodingException out of nextRecord and getBytePosition() goes wrong.

Passed the previous char explicitly and moved the counting ahead of the lastChar update. What happens for a pair split across two buffer reads is covered too since the first char still pairs against the saved lastChar. Added a regression test next to the existing multi-character-delimiter byte-position one.

### DIFF
--- a/src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
+++ b/src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
@@ -109,8 +109,10 @@ final class ExtendedBufferedReader extends UnsynchronizedBufferedReader {
 
     private long getEncodedCharLength(final char[] buf, final int offset, final int length) throws CharacterCodingException {
         int len = 0;
-        for (int i = offset; i < length; i++) {
-            len += getEncodedCharLength(buf[i]);
+        int previous = lastChar;
+        for (int i = offset; i < offset + length; i++) {
+            len += getEncodedCharLength(buf[i], previous);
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @digi-scrypt 
Please rebase on git master.
TY!

garydgregory:
Hello @digi-scrypt 

Would you please rebase on git master and resolve conflicts?

TY!

garydgregory:
@digi-scrypt ping 🔔

garydgregory:
Hello @digi-scrypt

Would you please rebase on git master and resolve conflicts?

TY!

garydgregory:
@digi-scrypt ping 🔔

garydgregory:
Hello @digi-scrypt

Would you please rebase on git master and resolve conflicts?

TY!

garydgregory:
@digi-scrypt 
Closing: I added the test to git master since it's green without changes to `main`.

PR #607, opened 2026-06-09
[CSV-327] Limit parser maxRows by produced records

OldTruckDriver:
Fixes CSV-327.

`CSVParser` used the next assigned record number to decide whether `maxRows` allowed another row. When parsing resumes with `CSVParser.Builder#setRecordNumber(...)`, this can make a parser with `maxRows` return no records before it has produced any rows.

This changes the iterator to track how many records it has produced and applies `maxRows` to that count instead of the assigned record number.

Tests:
- `mvn -q test`


garydgregory:
Jira ticket is https://issues.apache.org/jira/browse/CSV-327

garydgregory:
Thank you @OldTruckDriver , PR merged 🚀

PR #608, opened 2026-06-09
Clear escape delimiter buffer before peek in isEscapeDelimiter

rootvector2:
`isEscapeDelimiter()` peeks the next characters into the reused `escapeDelimiterBuf` look-ahead without clearing it first, so a truncated escaped multi-character delimiter at EOF is completed from the stale bytes of an earlier full escaped delimiter and the lexer appends a delimiter the input never contained. clearing the buffer before the peek mirrors the `delimiterBuf` reset added in `nextToken` for the plain multi-character delimiter path.

found while reading the partial-delimiter-at-EOF fix (CSV-324): the escaped-delimiter sibling was left unguarded.

for delimiter `[|]` and escape `!`, parsing `x![!|!]y![!|` returns `x[|]y[|]` before the change and the correct `x[|]y![!|` after. regression tests added in `LexerTest` and `CSVParserTest` (both fail without the one-line change).

- [ ] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.

### DIFF
--- a/src/main/java/org/apache/commons/csv/Lexer.java
+++ b/src/main/java/org/apache/commons/csv/Lexer.java
@@ -191,6 +191,7 @@ final class Lexer implements Closeable {
      * @throws IOException If an I/O error occurs.
      */
     boolean isEscapeDelimiter() throws IOException {
+        Arrays.fill(escapeDelimiterBuf, '\0');
         reader.peek(escapeDelimiterBuf);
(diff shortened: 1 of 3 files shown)


garydgregory:
Thank you @rootvector2 

PR merged 🚀

PR #610, opened 2026-06-15
Quote value starting with comment marker in minimal quote mode

rootvector2:
`printWithQuotes` (the default `MINIMAL` quote mode) decides whether to encapsulate a value by special-casing only the hard-coded `#` (`Constants.COMMENT`) as the first character, so a configured comment marker greater than `#` such as `;` is left unquoted. I hit this round-tripping printer output back through the parser: `CSVFormat.DEFAULT.builder().setCommentMarker(';').get()` prints `";id", "name"` as `;id,name`, and re-parsing with that same format reads the line as a comment and silently drops the whole record.

The encapsulation check is the right place to fix it, since that is already where the printer protects the default comment char, the delimiter, the quote and newlines. The change quotes a value whenever its first character equals the configured comment marker, so the output reads back as data.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -2454,10 +2454,11 @@ public final class CSVFormat implements Serializable {
                 }
             } else {
                 char c = charSeq.charAt(pos);
-                if (c <= Constants.COMMENT) {
+                if (c <= Constants.COMMENT || isCommentMarkerSet() && c == commentMarker.charValue()) {
                     // Some other chars at the start of a value caused the parser to fail, so for now
                     // encapsulate if we start in anything less than '#'. We are being conservative
-                    // by including the default comment char too.
+                    // by including the default comment char and any configured comment marker too,
+                    // which the parser would otherwise read back as a comment line.
                     quote = true;
                 } else {
                     while (pos < len) {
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @rootvector2 

Thank you for the PR.

Would you please update the new test to make sure that comments are properly printed. We want to make sure the new test shows the difference between printing a comment and printing a value that happens to start with the comment char.

Can you think of other cases to test?

rootvector2:
Done. The test now prints a real comment with `printComment` first, then the leading-marker value, so the output makes the difference explicit: `; a real comment` stays a comment line while `;comment-like` becomes `";comment-like"`.

Added a couple more cases too: the marker past the first character (`a;b`) is left unquoted since it can't start a comment, and a leading marker in a non-first column (`;c`) is still quoted. The round-trip check confirms the comment is dropped on read and both data records survive.

garydgregory (changes requested):
@rootvector2 
Please don't bother pushing to github until you've run `mvn` by itself to run and fix all build failures 🦺

rootvector2:
My fault. It was checkstyle `OperatorWrap` on the multi-line `assertEquals` concatenation, the `+` has to sit at the end of the line, not the start of the next. Fixed it and re-ran `mvn` by itself, build's green now with 0 violations. Amended into the test commit.

PR #611, opened 2026-06-17
Clear delimiter buffer before each peek in isDelimiter

rootvector2:
`isDelimiter` peeks the next characters into the reused `delimiterBuf` look-ahead buffer without clearing it, so a non-matching peek earlier in the same token leaves stale chars in the trailing positions and a truncated multi-character delimiter at EOF false-matches. With delimiter `[|]`, input `x[a][|` is split into `x[a]` and an empty field, because the earlier `[a]` peek left `]` in the buffer and the trailing `[|` (only two of the three delimiter chars present before EOF) matches against that stale `]`. CSV-324 cleared `delimiterBuf` once per `nextToken`, but `isDelimiter` peeks repeatedly within a token, so the reset has to happen before the peek, the same way `isEscapeDelimiter` already does it; the once-per-token clear is then redundant and dropped.

Found while auditing the multi-character delimiter look-ahead after CSV-324.

### DIFF
--- a/src/test/java/org/apache/commons/csv/CSVParserTest.java
+++ b/src/test/java/org/apache/commons/csv/CSVParserTest.java
@@ -1696,6 +1696,21 @@ class CSVParserTest {
         }
     }
 
+    /**
+     * A truncated multi-character delimiter at EOF must not be completed from the look-ahead buffer left dirty by an
+     * earlier non-matching peek in the same token.
+     */
+    @Test
+    void testPartialMultiCharacterDelimiterAtEOFAfterMismatch() throws IOException {
+        final CSVFormat format = CSVFormat.DEFAULT.builder().setDelimiter("[|]").get();
+        // The "[a]" peek leaves ']' in the look-ahead buffer; the trailing "[|" must not match "[|]".
+        try (CSVParser parser = format.parse(new StringReader("x[a][|"))) {
+            final CSVRecord record = parser.nextRecord();
+            assertEquals("x[a][|", record.get(0));
+            assertEquals(1, record.size());
+        }
+    }
+
     @Test
     void testProvidedHeader() throws Exception {
         final Reader in = new StringReader("a,b,c\n1,2,3\nx,y,z");
--- a/src/test/java/org/apache/commons/csv/LexerTest.java
+++ b/src/test/java/org/apache/commons/csv/LexerTest.java
@@ -433,6 +433,19 @@ class LexerTest {
         }
     }
 
+    /**
+     * A truncated multi-character delimiter at EOF must not be accepted by reusing the look-ahead buffer left dirty by an
+     * earlier non-matching peek in the same token (CSV-324 only cleared the buffer once per token).
+     */
+    @Test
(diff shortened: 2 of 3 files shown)


garydgregory (changes requested):
Hello @rootvector2 

Thank you for your PR.

Would you please add a test that shows why this is an issue only using the public parsing API?

rootvector2:
Added `testPartialMultiCharacterDelimiterAtEOFAfterMismatch` in `CSVParserTest` that drives this through the public parsing API. Parsing `x[a][|` with delimiter `[|]`: without the fix it splits into `x[a]` and an empty field, with the fix it stays a single field `x[a][|`. The test fails on `master` and passes with the patch.

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -1696,6 +1696,21 @@ void testPartialMultiCharacterDelimiterAtEOF() throws IOException {
...
 
+    /**
+     * A truncated multi-character delimiter at EOF must not be completed from the look-ahead buffer left dirty by an
+     * earlier non-matching peek in the same token.
+     */
+    @Test
+    void testPartialMultiCharacterDelimiterAtEOFAfterMismatch() throws IOException {
+        final CSVFormat format = CSVFormat.DEFAULT.builder().setDelimiter("[|]").get();
+        // The "[a]" peek leaves ']' in the look-ahead buffer; the trailing "[|" must not match "[|]".
+        try (CSVParser parser = format.parse(new StringReader("x[a][|"))) {
```
garydgregory, line 1707: @rootvector2 

Thank you for your update.

In the future, please make magic strings like `"x[a][|"` a local variable such that you never have to check that the assertion below is checking for a possibly different value.

src/test/java/org/apache/commons/csv/LexerTest.java
```diff
@@ -433,6 +433,19 @@ void testPartialMultiCharacterDelimiterAtEOF() throws IOException {
...
 
+    /**
+     * A truncated multi-character delimiter at EOF must not be accepted by reusing the look-ahead buffer left dirty by an
+     * earlier non-matching peek in the same token (CSV-324 only cleared the buffer once per token).
+     */
+    @Test
+    void testPartialMultiCharacterDelimiterAtEOFAfterMismatch() throws IOException {
+        final CSVFormat format = CSVFormat.DEFAULT.builder().setDelimiter("[|]").get();
+        // The "[a]" peek leaves ']' in the look-ahead buffer; the trailing "[|" must not match "[|]".
+        try (Lexer lexer = createLexer("x[a][|", format)) {
```
garydgregory, line 444: @rootvector2 

Thank you for your update.

In the future, please make magic strings like `"x[a][|"` a local variable such that you never have to check that the assertion below is checking for a possibly different value.

garydgregory:
@rootvector2 
Please see my comments for future PRs. Merged and thank you! 🚀

rootvector2:
makes sense, will pull magic strings like that into locals going forward. thanks for the merge.

PR #612, opened 2026-06-18
[CSV-328] Fix quoted null string after disabling quote

OldTruckDriver:
[CSV-328] Fix quoted null string after disabling quote

CSVFormat.Builder#setNullString(String) rebuilt quotedNullString by concatenating the nullable quoteCharacter field directly. When setQuote(null) was called before setNullString("NULL"), QuoteMode.ALL printed null as nullNULLnull instead of "NULL".

This change makes setNullString(String) and setQuote(Character) use the same quoted-null-string update logic, preserving the existing fallback to the default double quote when the quote character is disabled.

Tests cover both setter orders for setNullString("NULL") and setQuote(null).

Tests run:
- mvn -q -Dtest=org.apache.commons.csv.CSVFormatTest#testQuotedNullStringTracksQuoteCharacter test
- mvn -q -Dtest=org.apache.commons.csv.CSVFormatTest test
- mvn -q

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -780,8 +780,7 @@ public final class CSVFormat implements Serializable {
          */
         public Builder setNullString(final String nullString) {
             this.nullString = nullString;
-            this.quotedNullString = quoteCharacter + nullString + quoteCharacter;
-            return this;
+            return setQuotedNullString();
         }
 
         /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/csv/CSVFormat.java
```diff
@@ -806,9 +806,13 @@ public Builder setQuote(final Character quoteCharacter) {
                 throw new IllegalArgumentException("The quoteCharacter cannot be a line break");
             }
             this.quoteCharacter = quoteCharacter;
+            setQuotedNullString();
+            return this;
+        }
+
+        private void setQuotedNullString() {
```
garydgregory: @OldTruckDriver 
Please keep and use the fluent style.

garydgregory (changes requested):
Hello @OldTruckDriver 
Thank you for the PR, please see my comment.

src/main/java/org/apache/commons/csv/CSVFormat.java
OldTruckDriver: Done. Made setQuotedNullString() return the Builder and used return in the setters. Thanks

garydgregory:
Jira ticket is https://issues.apache.org/jira/browse/CSV-328
Merged 🚀 Thank you @OldTruckDriver

PR #614, opened 2026-06-19
Escape leading comment marker in printWithEscapes

rootvector2:
With an escape character set and no quote character (or `QuoteMode.NONE`), `CSVFormat.printWithEscapes` escapes the delimiter, `CR`, `LF`, the escape character and (since #609) the quote character, but never a comment marker. A value whose first character is the configured comment marker is written verbatim, so `CSVPrinter` emits a record that its own `CSVParser` reads back as a comment and silently drops. `CSVFormat.DEFAULT.builder().setQuote(null).setEscape('\').setCommentMarker(';').get()` prints `;foo` for the value `;foo`, and re-parsing that output yields zero records. Found round-tripping printer output back through the parser.

The escape condition is the right place to fix it, next to where the delimiter, escape char and quote char are already handled. The change escapes the comment marker when it is the first character of the value, in both `printWithEscapes` overloads (`CharSequence` and `Reader`). This is the escape-mode counterpart to #610, which protected the comment marker only in the `MINIMAL` quoting path and left the escape paths out.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -2329,12 +2329,16 @@ public final class CSVFormat implements Serializable {
         final char escape = getEscapeChar();
         final boolean quoteSet = isQuoteCharacterSet();
         final char quote = quoteSet ? getQuoteCharacter().charValue() : 0;
+        final boolean commentMarkerSet = isCommentMarkerSet();
+        final char commentChar = commentMarkerSet ? commentMarker.charValue() : 0; // Explicit unboxing is intentional
         while (pos < end) {
             char c = charSeq.charAt(pos);
             final boolean isDelimiterStart = isDelimiter(c, charSeq, pos, delimArray, delimLength);
             final boolean isCr = c == Constants.CR;
             final boolean isLf = c == Constants.LF;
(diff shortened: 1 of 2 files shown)


garydgregory:
@rootvector2 

The build is broken.
Please run `mvn` with no arguments BEFORE you push.
This will avoid wasting time and CI resources.

rootvector2:
ran `mvn` here and the only failure is `testGetBytePositionMultiCharacterDelimiterWithSupplementaryCharacter`, which fails on `master` too, not from this PR. the `Refactor delimiter in test` commit (a1cf4f2) set the expected value to `"a" + delimiter + "b\n".getBytes(UTF_8).length`. `.length` binds before `+`, so that evaluates to the string `"ax😀2"` rather than the byte count `8`; the prior `"ax😀b\n".getBytes(UTF_8).length` was correct. my change and its two tests are green once that line is restored. want me to drop the one-line fix into this PR, or will you patch master?

garydgregory:
You're right @rootvector2 ! Please rebase on git `master`.

rootvector2:
rebased onto `master`, so the revert of `a1cf4f2` is in the branch now and the byte-position test is green again. ran `mvn` with no args locally, full build passes.

garydgregory:
TY @rootvector2 , merged 🚀

PR #615, opened 2026-06-20
Skip byte counting at EOF in ExtendedBufferedReader.read

rootvector2:
`ExtendedBufferedReader.read()` feeds the `EOF` sentinel (`-1`) to `getEncodedCharLength` when `trackBytes` is on, narrowing it to `(char) -1` (`U+FFFF`) and pushing it through the charset encoder. A single-byte charset like `ISO-8859-1` or `US-ASCII` cannot encode `U+FFFF`, so byte tracking throws `UnmappableCharacterException` at end of input on an otherwise valid file; with `UTF-8`/`UTF-16` it silently adds a few phantom bytes to the count. Found while running `getBytePosition` with a non-UTF charset; the `read(char[], off, len)` overload already guards this with `len > 0`, so this mirrors that guard for the single-char path.


garydgregory:
Good catch

garydgregory:
Good catch @rootvector2 , merged 🚀

PR #616, opened 2026-06-21
Keep quoted empty trailing field with trailingDelimiter

rootvector2:
`addRecordValue` drops the empty last field whenever the format's `trailingDelimiter` is enabled, checking only that the trimmed value is empty. A quoted empty field at the end of a record is a real value, not a trailing delimiter, so `a,b,""` parsed with `setTrailingDelimiter(true)` came back as two values `[a, b]` instead of three `[a, b, ""]`. The unquoted case `a,b,` is a genuine trailing delimiter and still collapses to two fields. Fixed by only dropping the field when the token is unquoted; `reusableToken.isQuoted` is already tracked and used by `handleNull`.

Found while round-tripping records that contain an explicit empty quoted field.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -883,6 +883,16 @@ public final class CSVFormat implements Serializable {
         /**
          * Sets whether to add a trailing delimiter.
          *
+         * <p>
+         * When writing, a delimiter is appended after the last value of each record. When reading, the empty field
+         * that such a trailing delimiter produces is dropped so the output round-trips back to the original record;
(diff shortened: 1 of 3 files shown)


garydgregory (changes requested):
Thank you for your update.
Would you please update the Javadoc for "trailing delimiter" to explain the behavior. Especially contrasting with "trailing data".
TY!

rootvector2:
Done. Expanded the Javadoc on `setTrailingDelimiter`/`getTrailingDelimiter` to cover the write side (delimiter appended after the last value) and the read side (the empty field it produces is dropped, but a quoted empty trailing field `""` is kept). Also noted it's unrelated to `trailingData`, which is about characters after a closing quote.

garydgregory:
TY @rootvector2 , merged 🚀

PR #620, opened 2026-07-01
Quote value ending with multi-character delimiter prefix

rootvector2:
`printWithQuotes` in MINIMAL quote mode does not encapsulate a value whose tail is a straddling prefix of a multi-character delimiter, so the printed record cannot be read back. With delimiter `||`, a field ending in `|` such as `a|` is written unquoted, and the record `a|||b` is then parsed by `CSVParser` as `[a, |b]` because the greedy lexer matches the delimiter one character early at the value/delimiter boundary. The existing scan only quotes when the full delimiter occurs inside the value; a trailing partial delimiter slips through. Fixed by `endsWithDelimiterPrefix`, which quotes the value when appending the delimiter after it would create a boundary-straddling delimiter match. Single-character delimiters and values that contain the delimiter start without a straddling suffix (for example delimiter `[|]`, value `a[`) are unaffected.

Found while auditing multi-character delimiter round-trips through the printer.

### DIFF
--- a/src/main/java/org/apache/commons/csv/CSVFormat.java
+++ b/src/main/java/org/apache/commons/csv/CSVFormat.java
@@ -2098,6 +2098,33 @@ public final class CSVFormat implements Serializable {
     }
 
     /**
+     * Tests whether appending the delimiter after {@code charSeq} would let the parser match the delimiter starting inside the value. This happens with a
+     * multi-character delimiter when the value ends with a straddling prefix of it (for delimiter {@code ||}, a value ending in {@code |} followed by the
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @rootvector2 
Do supplementary characters need to be accounted for in the new code?

rootvector2:
No special handling needed. `endsWithDelimiterPrefix` compares `char` by `char`, same as `isDelimiter` and the lexer's delimiter match, so a supplementary code point (a surrogate pair) is matched symmetrically on both the print and parse sides. Added `testQuoteValueEndingWithSupplementaryDelimiterPrefix` with a two-clef (U+1D11E) delimiter and a value ending in a single clef to prove the straddle is quoted and round-trips.

garydgregory:
TY @rootvector2 , merged 🚀

PR #622, opened 2026-07-13
Exclude byte-order mark from ExtendedBufferedReader byte counting

rootvector2:
`getEncodedCharLength` re-encodes each character with the configured `CharsetEncoder` to count its bytes, but the convenience `encode` call re-emits a byte-order mark on every invocation for charsets like `UTF-16`. Parsing a UTF-16 stream with `setTrackBytes(true)`, for example an Excel Unicode export, counts every code unit as 4 bytes instead of 2, so `CSVRecord.getBytePosition()` grows at twice the true rate. Found by comparing reported byte offsets against the encoded prefix length across charsets.

Measure the per-call BOM once and subtract it from each encoded length. The count already lives in `getEncodedCharLength`, so keeping the adjustment there covers both the single-char and array read paths, and it stays a no-op for charsets that emit no BOM (`UTF-8`, `UTF-16LE`/`BE`, single-byte).

### DIFF
--- a/src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
+++ b/src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
@@ -66,6 +66,9 @@ final class ExtendedBufferedReader extends UnsynchronizedBufferedReader {
     /** Encoder for calculating the number of bytes for each character read. */
     private final CharsetEncoder encoder;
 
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @rootvector2 , merged 🚀 
Would you please add a similar test for UTF-32 BE/LE?

PR #623, opened 2026-07-16
Skip trailing delimiter on comment lines in printComment

rootvector2:
`printComment` closes each comment line with `println()`, which appends the trailing delimiter when `setTrailingDelimiter(true)` is set. So a multi-line comment prints `# line1,` / `# line2,` and the parser reads that delimiter back as part of the comment text. Found round-tripping comments through a printer with the trailing delimiter enabled.

Close comment lines with the record separator only. The trailing delimiter is documented as following the last value of a record, and a comment is not a record, so the record and header paths keep calling `println()` and stay unchanged.


garydgregory:
Good find @rootvector2 ; merged 🚀

PR #625, opened 2026-07-20
Fill the lookahead buffer on short reads from a chunked source

saleemno1:
Thanks for maintaining this! `ExtendedBufferedReader.peek(char[])` and `read(char[], int, int)` inherit a buffered read that returns as soon as the source reports it is not ready, so a `Reader` backed by a socket or a pipe hands back a short read and the delimiter lookahead is matched against a half-filled buffer.

Two things go wrong once the buffer is short:

- `Lexer.isDelimiter` / `isEscapeDelimiter` compare every slot, so the still-zeroed tail makes a multi-character delimiter that is really present look like ordinary text. Parsing `a[|]b` with delimiter `[|]` yields one field `a[|]b` from a chunked reader and two fields `a`, `b` from a `StringReader` for the same bytes, which is a parser differential if the split feeds any filtering or routing decision.
- `CSVFormat.printWithEscapes(Reader, Appendable)` builds its lookahead the same way, so printing the single value `x[|]y` from a chunked reader writes the delimiter unescaped and the record reads back as two fields.

Both call sites share the same reader, so the loop lives there rather than in each caller. Added a regression test at the reader level and one at the parser level; both fail without the change. `mvn` is green including the coverage gate.

### DIFF
--- a/src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
+++ b/src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
@@ -225,6 +225,37 @@ final class ExtendedBufferedReader extends UnsynchronizedBufferedReader {
         super.mark(readAheadLimit);
     }
 
+    /**
+     * Fills {@code array} with the characters that follow the current position without consuming them.
+     * <p>
+     * Overridden because the inherited implementation stops at the first short read, which leaves the tail of the array holding stale content when the source
+     * delivers data in chunks. Callers compare the whole array against a multi-character delimiter, so a partial fill makes them miss a delimiter that is
+     * really there.
+     * </p>
+     *
+     * @param array the buffer to fill.
+     * @return the number of characters peeked, or {@link IOUtils#EOF} at the end of the stream.
+     * @throws IOException If an I/O error occurs.
+     */
+    @Override
+    public int peek(final char[] array) throws IOException {
+        final int length = array.length;
+        if (length == 0) {
+            return 0;
+        }
+        super.mark(length);
+        int len = 0;
+        while (len < length) {
+            final int more = super.read(array, len, length - len);
+            if (more == EOF) {
+                break;
+            }
+            len += more;
+        }
+        super.reset();
+        return len == 0 ? EOF : len;
+    }
+
     @Override
     public int read() throws IOException {
         final int current = super.read();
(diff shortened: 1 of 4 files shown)


src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -1709,6 +1709,47 @@ void testParseWithDelimiterStringWithEscape() throws IOException {
         }
     }
 
+    @Test
+    void testParseWithDelimiterStringFromChunkedReader() throws IOException {
+        // A reader that hands out one character at a time and never reports itself ready, like a socket or pipe.
+        final Reader chunked = new Reader() {
```
garydgregory: If this is the same as the behavior of the `ChunkedReader` in the other test, then refactor for reuse.

src/test/java/org/apache/commons/csv/ExtendedBufferedReaderTest.java
```diff
@@ -240,4 +240,72 @@ void testReadLookahead2() throws Exception {
...
+            assertEquals(2, br.peek(peeked));
+            assertEquals('a', peeked[0]);
+            assertEquals('b', peeked[1]);
+        }
+    }
+
+    /**
+     * A reader that returns one character per call and never reports itself ready, like a socket or pipe that delivers data in chunks.
+     */
+    private static final class ChunkedReader extends java.io.Reader {
```
garydgregory: Would it be simpler to extend `StringReader` and override those methods?

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
```diff
@@ -244,7 +275,16 @@ public int read(final char[] buf, final int offset, final int length) throws IOE
         if (length == 0) {
             return 0;
         }
-        final int len = super.read(buf, offset, length);
+        int len = super.read(buf, offset, length);
+        // The underlying buffered reader stops early once the source reports it is not ready, so a stream that delivers data in chunks (a socket or a pipe)
+        // yields a short read. Callers match multi-character sequences against this buffer, so keep reading until it is full or the source is exhausted.
```
garydgregory, line 280: Isn't there still a risk that even filling to the end of the buffer breaks up multi-character sequences?

garydgregory (changes requested):
Hello @saleemno1 
Please see my questions and comments.

src/test/java/org/apache/commons/csv/ExtendedBufferedReaderTest.java
saleemno1: Done. It's now a small `ChunkedReader extends StringReader` that overrides `read(char[], int, int)` to hand back one char and `ready()` to return false, which is a lot less code than the raw `Reader`.

src/test/java/org/apache/commons/csv/CSVParserTest.java
saleemno1: Right, they were the same thing. Pulled it out into a package-private `ChunkedReader` in the test tree so both this test and ExtendedBufferedReaderTest use it.

src/main/java/org/apache/commons/csv/ExtendedBufferedReader.java
saleemno1: No, because the callers size the array to the exact sequence they're matching. Lexer allocates `delimiterBuf` as `delimiter.length - 1` and `escapeDelimiterBuf` as `2 * delimiter.length - 1`, so once the array is full it holds the whole tail of the delimiter (or escaped delimiter), never a piece of it. Same for `CSVFormat.printWithEscapes`, which peeks `delimiter.length - 1`.

The only way the loop exits short now is EOF, and then the remaining characters genuinely don't exist, so no complete sequence can be there to miss. Before the change the loop could exit short just because the source wasn't ready yet, which is the case that broke.

PR #626, opened 2026-07-21
Keep an escaped value that equals the null string

saleemno1:
`CSVParser.handleNull` compares the fully unescaped token against the format's null string, so a value that really is `\N` and that the printer correctly wrote as `\\N` for `MYSQL`, `POSTGRESQL_TEXT`, `POSTGRESQL_CSV` and `ORACLE` comes back as `null`; the lexer now records whether an escape sequence was actually translated, and only an untranslated token can be the null marker.

### DIFF
--- a/src/test/java/org/apache/commons/csv/CSVParserTest.java
+++ b/src/test/java/org/apache/commons/csv/CSVParserTest.java
@@ -587,7 +587,7 @@ class CSVParserTest {
     @ParameterizedTest
     @EnumSource(value = CSVFormat.Predefined.class, names = { "MySQL", "PostgreSQLCsv", "PostgreSQLText", "Oracle" })
     void testEscapedNullStringIsAValue(final CSVFormat.Predefined predefined) throws Exception {
-        // For formats whose null string is "\\N" (e.g., MySQL, PostgreSQL Text, Oracle), 
+        // For formats whose null string is "\\N" (e.g., MySQL, PostgreSQL Text, Oracle),
         // a literal value "\\N" must be written as "\\\\N" so it is not read back as null.
         final CSVFormat format = predefined.getFormat();
         final StringWriter writer = new StringWriter();
--- a/src/test/java/org/apache/commons/csv/CSVParserTest.java
+++ b/src/test/java/org/apache/commons/csv/CSVParserTest.java
@@ -584,19 +584,19 @@ class CSVParserTest {
         }
     }
 
-    @Test
-    void testEscapedNullStringIsAValue() throws Exception {
+    @ParameterizedTest
+    @EnumSource(value = CSVFormat.Predefined.class, names = { "MySQL", "PostgreSQLCsv", "PostgreSQLText", "Oracle" })
+    void testEscapedNullStringIsAValue(final CSVFormat.Predefined predefined) throws Exception {
         // "\N" is the MySQL and PostgreSQL null marker, "\\N" is the value "\N", which is what the printer writes for it.
-        for (final CSVFormat format : new CSVFormat[] { CSVFormat.MYSQL, CSVFormat.POSTGRESQL_TEXT, CSVFormat.ORACLE }) {
-            final StringWriter writer = new StringWriter();
-            try (CSVPrinter printer = new CSVPrinter(writer, format)) {
-                printer.printRecord("\\N", null);
-            }
-            try (CSVParser parser = CSVParser.parse(writer.toString(), format)) {
-                final CSVRecord record = parser.nextRecord();
-                assertEquals("\\N", record.get(0), format.toString());
-                assertNull(record.get(1), format.toString());
-            }
+        final CSVFormat format = predefined.getFormat();
+        final StringWriter writer = new StringWriter();
+        try (CSVPrinter printer = new CSVPrinter(writer, format)) {
+            printer.printRecord("\\N", null);
+        }
+        try (CSVParser parser = CSVParser.parse(writer.toString(), format)) {
+            final CSVRecord record = parser.nextRecord();
+            assertEquals("\\N", record.get(0));
+            assertNull(record.get(1));
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -584,6 +584,22 @@ void testEndOfFileBehaviorExcel() throws Exception {
         }
     }
 
+    @Test
+    void testEscapedNullStringIsAValue() throws Exception {
+        // "\N" is the MySQL and PostgreSQL null marker, "\\N" is the value "\N", which is what the printer writes for it.
+        for (final CSVFormat format : new CSVFormat[] { CSVFormat.MYSQL, CSVFormat.POSTGRESQL_TEXT, CSVFormat.ORACLE }) {
```
garydgregory: Use a JUnit `@ParameterizedTest` instead of a loop.

garydgregory (changes requested):
Hello @saleemno1 
Please see my comment.

src/test/java/org/apache/commons/csv/CSVParserTest.java
saleemno1: Done, it's an @ParameterizedTest with @EnumSource now. I also folded PostgreSQLCsv into the list since it has the same \N null string and was missing from the loop.

src/test/java/org/apache/commons/csv/CSVParserTest.java
```diff
@@ -584,6 +584,22 @@ void testEndOfFileBehaviorExcel() throws Exception {
...
     }
 
+    @ParameterizedTest
+    @EnumSource(value = CSVFormat.Predefined.class, names = { "MySQL", "PostgreSQLCsv", "PostgreSQLText", "Oracle" })
+    void testEscapedNullStringIsAValue(final CSVFormat.Predefined predefined) throws Exception {
+        // "\N" is the MySQL and PostgreSQL null marker, "\\N" is the value "\N", which is what the printer writes for it.
+        final CSVFormat format = predefined.getFormat();
+        final StringWriter writer = new StringWriter();
+        try (CSVPrinter printer = new CSVPrinter(writer, format)) {
+            printer.printRecord("\\N", null);
```
garydgregory: Refactor  the magic string `\\N` into a constant or a local variable with a name that explains its intent. Since there are 2 such instances in this method, it will also make it obvious that both magic strings are indeed intended to be the same.

garydgregory (changes requested):
Hello @saleemno1 
Thank you for your update. Please see my comment.

garydgregory:
@saleemno1 Ping on https://github.com/apache/commons-csv/pull/626#discussion_r3630790096 👋

garydgregory (changes requested):
@saleemno1 
Please review the copilot comments and at least improve the code comments.
TY!

garydgregory:
@saleemno1 ping 🔔

src/test/java/org/apache/commons/csv/CSVParserTest.java
saleemno1: Done. Pulled it into a local `valueEqualToNullString` so both spots reference the same value.

saleemno1:
Also went through the Copilot notes: reworded the handleNull comment so it no longer implies null is always written unquoted (QuoteMode.ALL can quote it, it just never escapes it), parenthesized the isEscaped/strict-quote condition, and tightened the expectNulls javadoc and the test comment to match what actually happens for each format (PostgreSQL CSV uses an empty null string, so its \N field round trips as the value too).

garydgregory:
@saleemno1 
Please rebase on git master and try again.

garydgregory:
Converting to draft as I not convinced this is a bug instead of a behavior we haven't defined.

saleemno1:
Rebased on master (had one trivial conflict where both branches added a private test helper, kept both) and pushed. Full CSVParserTest/CSVPrinterTest suite is green.

On whether it's a bug: the way I see it, the printer already treats these two cases as distinct on the write side. For nullString `\N`, a real null is written verbatim as `\N`, but a field whose value is literally `\N` gets escaped to `\\N` precisely so it doesn't collide with the null marker. The parser was then unescaping `\\N` back to `\N` and mapping it to null, so `printRecord("\\N", null)` doesn't survive a print/parse round trip: the first field silently becomes null. So it's less about defining new behavior and more that parse wasn't the inverse of print for this one case.

That said, it's your call whether that round trip is something the library promises. If you'd rather leave it undefined I'm happy to close this out; if it's worth fixing, this keeps the existing null handling intact and only diverges on the escaped case.

saleemno1:
Fair to want that nailed down. The reason I read it as a bug and not just undefined: it is a round trip through the library's own printer and parser with the same predefined format. For MySQL/PostgreSQLText/Oracle, `printRecord("\N", null)` writes the value as `\\N` and the null as `\N`, so the printer already escapes the value specifically to keep it distinct from the null marker. On the way back the parser dropped that distinction and both fields came back as null, so a real value silently became null.

So the two halves of the library disagree on the same format, and the escape the printer itself emitted is exactly what tells them apart. That is all the patch keys off. `testEscapedNullStringIsAValue` in the PR is that round trip and fails on master today.

Totally your call on scope, but that is why it looks like a correctness gap rather than open behavior to me.

garydgregory:
@saleemno1 Please rebase on git master to pick up the fix for the Javadoc break. In meantime, I'll review...

garydgregory:
@saleemno1 ping 🔔

garydgregory:
@saleemno1 Please rebase on git master to pick up the fix for the Javadoc break.

garydgregory:
@saleemno1 Please rebase on git master to pick up the fix for the Javadoc break.

garydgregory:
Closing: No reply, PR fixed locally and applied to git master, credited in `changes.xml`.

PR #630, opened 2026-08-01
close the input when createHeaders throws in the CSVParser constructor

rootvector2:
`createHeaders()` runs inside the `CSVParser` constructor, after the lexer already owns the reader returned by `builder.getReader()`, so a duplicate or missing header name throws before the caller receives a `Closeable` and nothing is left to close the input with. For `parse(Path)`, `parse(File)`, `parse(URL)` and the `Path`/`File`/`URI` builder forms the library opened that stream itself, so the descriptor or connection stays open for the lifetime of the JVM. Found while checking what happens to the handle when a header fails validation.

The constructor is the layer that took ownership and every entry point funnels its reader through it, so closing there covers all of them; a close failure is attached as suppressed so the original error still surfaces.


garydgregory:
Hello @rootvector2 

This one feels wrong to me. 

In git master, please find the new `CSVParserTest.testTryWithResourcesParse*`

- `testTryWithResourcesParseInputStreamWhenHeaderIsInvalid`
- `testTryWithResourcesParseReaderWhenHeaderIsInvalid`

It is the responsibility of the code that allocates a resource to close it.

If the the resource is managed with a try-with-resources block (see the new tests), the resource is closed for this use case. I switched the PR to draft for this reason.

Please let me know how you see it.

TY!

rootvector2:
Agreed on caller-supplied readers; the new `testTryWithResourcesParse*` tests cover that case, so I dropped that part.

The case I was after is narrower: `parse(Path)` and `parse(URL)` allocate the stream themselves via `Files.newInputStream(path)` and `url.openStream()`. When `createHeaders()` throws, the caller never gets a parser back, so there is nothing to put in a try-with-resources block and the stream the library opened is unreachable. Applying the same rule, the method that allocated it now closes it on failure; caller-supplied readers and streams are untouched. Reworked the PR that way and rebased onto master. Your new tests still pass, and `testParseUrlClosesInputStreamWhenHeaderIsInvalid` fails without the change.

garydgregory:
@rootvector2 
I brought this in with a refactored implementation but a better test for Path would be good, or not, up to you ;)

garydgregory:
@rootvector2 ping 🔔

rootvector2:
Right, the temp file test showed nothing: deleting an open file succeeds on POSIX, so the only observable effect was the throw. Reworked the PR into just a better `Path` test on top of your refactoring. `Files.newInputStream` dispatches through `path.getFileSystem().provider()`, so mocking that chain lets the test hand `parse(Path)` a close-recording stream, same shape as the URL test. It fails on master with the close in `parse(IOSupplier)` removed and passes as is, on any platform. Full default `mvn` build is green. Thanks for bringing the fix in.

